PR #27, opened 2021-10-27
adding prompt for pw but not supplied

kenvanhyning:
Adding prompt for missing pw to match legacy version of sqlcmd.  Changed the behavior of setting the params passed into ConnectDb() so please take a close look there.  Attempting to clarify behavior of the optional params and corresponding matching env vars.


shueybubbles:
no longer needed

PR #30, opened 2021-11-02
Stuartpa/add dash questionmark

stuartpa:
The first thing I did with the new sqlcmd is run sqlcmd -?, which doesn't display any useful help.  Add -? as a equiv of --help for back compat.


shueybubbles:
The online diff shows lots of whitespace difference. Do you know what that's coming from?

shueybubbles (approved):
Pls figure out what the whitespace change is in case we are using conflicting editors/linters or something.

PR #51, opened 2022-03-07
Implement connect command and password prompt

### LINKED ISSUE #50: sqlcmd doesn't prompt for password when required
If the `SQLCMDPASSWORD` environment variable isn't set, and sqlcmd is running in interactive mode, it should prompt the user for the password if required.

shueybubbles:
`:Connect` implementation uses the same command line parser as `main`. I've added a `-G` to enable AAD auth. 
I've also created a `Console` interface to enable future replacement of `readline` with something more modern.
Fixes #50


kenvanhyning (approved):
nice refactoring!

PR #53, opened 2022-03-09
Replace console module with liner

### LINKED ISSUE #31: Pasting T/SQL into interactive console is very slow (and omits characters sometimes)
When I paste a multi-line query (for example [1]) into the interactive console using the ODBC version of sqlcmd, the paste action completes within .1 of a second.

When I paste the same multi-line query into the go-mssql interactive console it takes a few seconds.  If I paste (as a single paste) the same query repeated 10 times into the the interactive console, it takes over a minute (and gets slower and slower).

Also, strangely, sometimes the two 'd' in the string 'Dedicated Admin Connection' are missed out. i.e. what is pasted in is 'Deicated  Amin Connection', and the 'f' in the line 'from sys.endpoints as ep' is omitted.  

[1]
select 
CASE
WHEN es.session_id= @@SPID THEN 'you are :)'
ELSE es.host_name
END AS Who_is_running_DAC,
es.original_login_name,
es.session_id,
es.login_time,
es.status
from sys.endpoints as ep
join sys.dm_exec_sessions es on
ep.endpoint_id=es.endpoint_id
where ep.name='Dedicated Admin Connection'

shueybubbles:
Fixes #31 
Bonus feature - tab completion of SQL keywords!
Unfortunately, the keyword list is hard coded based on recent sqlparser updates, as I know of no way to discover a keyword list from SQL itself.
There are other console input modules in the Golang universe but this one appears to be maintained and not-too-complicated.


pkg/console/complete.go
```diff
@@ -0,0 +1,245 @@
...
+			}
+			return lines
+		}
+	}
+	return []string{}
+}
+
+// mergeline appends keyword to line starting at index idx
+// It matches the case of the current character in the line
+func mergeLine(line string, keyword string, idx int) string {
```
shueybubbles, line 52: Question - should we add a space after the keyword? I find it natural to hit the spacebar myself after the tab cycles to the word I am looking for but maybe that's just me.

pkg/console/complete.go
```diff
@@ -0,0 +1,245 @@
...
+	b.Write([]byte(line[:idx]))
+	if !upcase {
+		b.WriteString(strings.ToLower(keyword))
+	} else {
+		b.WriteString(keyword)
+	}
+	return b.String()
+}
+
+var keywords = []string{
```
shueybubbles, line 64: Eventually we could enable users to override the keyword list with their own file.  We could also add auto complete for commands.

pkg/console/complete.go
dzsquared: any completion functionality I'm familiar with also does not add the space after the keyword, I do.  This is especially important for when you might accidentally complete to a keyword that is a component of another keyword.

pkg/console/complete.go
dzsquared: a common format for keywords is the plist file - would be interesting for them to be imported from there.
we might even be able to reuse https://github.com/microsoft/vscode-mssql/blob/main/syntaxes/SQL.plist

dzsquared (approved):

PR #59, opened 2022-03-20
Add version resource to Windows binaries

shueybubbles:
When we have an icon I will add that as well. The version data matches the ODBC sqlcmd.exe.
The version numbers will use the version tag and commit id on the branch.


build/azure-pipelines/build-common.yml
```diff
@@ -12,6 +12,23 @@ steps:
...
+  displayName: 'Go install go-winres'
+  inputs:
+    command: 'custom'
+    customCommand: 'install'
+    arguments: 'github.com/tc-hib/go-winres@latest'
+    workingDirectory: '$(Build.SourcesDirectory)/cmd/sqlcmd'
+  env:
+      GOBIN: $(Build.SourcesDirectory)
+
+- task: CmdLine@2
```
shueybubbles, line 27: I tried to use `go generate` instead but couldn't get ADO `task.prependpath` to add GOBIN to the path successfully on Windows.

stuartpa (approved):

PR #60, opened 2022-03-22
Support for reading UTF16 and UTF-8 script files. Support for processing files with long lines.

simonoff:
Support for big files.
The main issue that we have a script inside file around 600Mb.

### DIFF
--- a/pkg/sqlcmd/sqlcmd.go
+++ b/pkg/sqlcmd/sqlcmd.go
@@ -21,6 +21,8 @@ import (
 	mssql "github.com/denisenkom/go-mssqldb"
 	"github.com/denisenkom/go-mssqldb/msdsn"
 	"github.com/golang-sql/sqlexp"
+	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/transform"
 )
 
 var (
@@ -31,6 +33,8 @@ var (
 	// ErrCtrlC indicates execution was ended by ctrl-c or ctrl-break
 	ErrCtrlC = errors.New(WarningPrefix + "The last operation was terminated because the user pressed CTRL+C")
 )
+
+const maxLineBuffer = 2 * 1024 * 1024 // 2Mb
 
 // Console defines methods used for console input and output
 type Console interface {
@@ -112,7 +116,6 @@ func (s *Sqlcmd) Run(once bool, processAll bool) error {
 		} else {
 			cmd, args, err = s.batch.Next()
 		}
-
 		if err != nil {
 			if err == io.EOF {
 				if s.batch.Length == 0 {
@@ -283,13 +286,13 @@ func (s *Sqlcmd) IncludeFile(path string, processAll bool) error {
 		return err
 	}
 	defer f.Close()
-	info, _ := f.Stat()
-	var maxSize int
 	b := s.batch.batchline
-	scanner := bufio.NewScanner(f)
-	maxSize = int(info.Size())
-	buffer := make([]byte, 0, maxSize)
-	scanner.Buffer(buffer, maxSize)
+	win16be := unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)
+	utf16bom := unicode.BOMOverride(win16be.NewDecoder())
+	unicodeReader := transform.NewReader(f, utf16bom)
+	scanner := bufio.NewScanner(unicodeReader)
+	buf := make([]byte, maxLineBuffer)
+	scanner.Buffer(buf, maxLineBuffer)
 	curLine := s.batch.read
 	s.batch.read = func() (string, error) {
 		if !scanner.Scan() {
--- a/pkg/sqlcmd/sqlcmd.go
+++ b/pkg/sqlcmd/sqlcmd.go
@@ -283,8 +283,13 @@ func (s *Sqlcmd) IncludeFile(path string, processAll bool) error {
 		return err
 	}
 	defer f.Close()
+	info, _ := f.Stat()
+	var maxSize int
 	b := s.batch.batchline
 	scanner := bufio.NewScanner(f)
+	maxSize = int(info.Size())
+	buffer := make([]byte, 0, maxSize)
+	scanner.Buffer(buffer, maxSize)
 	curLine := s.batch.read
 	s.batch.read = func() (string, error) {
 		if !scanner.Scan() {
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/peterh/liner v1.2.2
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/text v0.3.6
 )
 
 replace github.com/denisenkom/go-mssqldb => github.com/shueybubbles/go-mssqldb v0.10.1-0.20220317022252-fafb9d92e469


shueybubbles:
Pls add a test that fails without this fix. It should be fairly straightforward to create a temp file big enough for the repro.

pkg/sqlcmd/sqlcmd.go
```diff
@@ -283,8 +283,13 @@ func (s *Sqlcmd) IncludeFile(path string, processAll bool) error {
 		return err
 	}
 	defer f.Close()
+	info, _ := f.Stat()
+	var maxSize int
 	b := s.batch.batchline
 	scanner := bufio.NewScanner(f)
+	maxSize = int(info.Size())
+	buffer := make([]byte, 0, maxSize)
```
shueybubbles, line 291: The buffer is meant to be the size of the largest token; using the file size seems excessive. Isn't allocation of a huge buffer like this more likely to fail on machines under stress? Or is your issue that the query is a single huge line?

shueybubbles (changes requested):
I am curious if there's a further restriction on the largest query line length in the driver or TDS itself that's smaller than the length of your big file. We shouldn't try to allocate a line longer than what we can send to the driver.

For files whose size is larger than scanner.MaxScanTokenSize, might it be worthwhile to use a custom `SplitFunc` instead and break the file into lines ourselves? We could use a fixed size buffer and a `strings.Builder`.

Or we could take a small performance hit and binary search the buffer size between {MaxScanTokenSize, file size} and handle the panic/error from Scan appropriately in a loop.
[edit] varying the buffer size probably requires keeping track of how far we made it in the file before the panic/error and avoiding re-reads of the same portion.

shueybubbles:
oh yes, and thanx a ton for being our first public PR submitter!

pkg/sqlcmd/sqlcmd.go
simonoff: yeah, i already changed such logic. As now i'm set it to 2MB size. 
Also, i  have found an issue with using a sql script on macos/linux inside docker. I'm still working on the fix and will push updates as i will finish it.

simonoff:
@shueybubbles I'm new to MS SQL completely but I'm following my usual workflow with PostgreSQL and now i think will do more in such case.

pkg/sqlcmd/sqlcmd.go
```diff
@@ -284,7 +287,12 @@ func (s *Sqlcmd) IncludeFile(path string, processAll bool) error {
 	}
 	defer f.Close()
 	b := s.batch.batchline
-	scanner := bufio.NewScanner(f)
+	win16be := unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)
```
shueybubbles, line 290: Please add an entry to the readme describing the supported text encodings of the files.  Having test files of each type will be great!

  The ODBC sqlcmd claims 
`sqlcmd automatically recognizes both big-endian and little-endian Unicode input files. If the -u option has been specified, the output will always be little-endian Unicode.`

We haven't implemented `-u` for output yet.

simonoff: aha... so need to add an option... I will add the tests today.

shueybubbles: if you are fine with UTF8 output you can skip implementing `-u` for now to unblock your input needs.

XC-:
Any progress on this one?

Based on the changes requested, I guess this is only for big endian, what about UTF-16LE then?

pkg/sqlcmd/sqlcmd.go
```diff
@@ -284,7 +287,12 @@ func (s *Sqlcmd) IncludeFile(path string, processAll bool) error {
 	}
 	defer f.Close()
 	b := s.batch.batchline
-	scanner := bufio.NewScanner(f)
+	win16be := unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM)
+	utf16bom := unicode.BOMOverride(win16be.NewDecoder())
+	unicodeReader := transform.NewReader(f, utf16bom)
+	scanner := bufio.NewScanner(unicodeReader)
+	buf := make([]byte, maxLineBuffer)
+	scanner.Buffer(buf, maxLineBuffer)
```
shueybubbles, line 295: I tried making this change locally and it breaks reading utf8 input.

simonoff: aha... thx for info. I'm now looking the way how to running tests on my M1. Could be that i will try once more time on this weekend.

shueybubbles: image

Shows the LE BOM, and in the debugger I see `bomOverride` read it correctly and set `d.current` to a `utf16le.NewDecoder()`

```go
func (d *bomOverride) Transform(dst, src []byte, atEOF bool) (nDst, nSrc int, err error) {
	if d.current != nil {
		return d.current.Transform(dst, src, atEOF)
	}
	if len(src) < 3 && !atEOF {
		return 0, 0, transform.ErrShortSrc
	}
	d.current = d.fallback
	bomSize := 0
	if len(src) >= 2 {
		if src[0] == 0xFF && src[1] == 0xFE {
====>		d.current = utf16le.NewDecoder()
			bomSize = 2
		} else if src[0] == 0xFE && src[1] == 0xFF {
   	d.current = utf16be.NewDecoder()
			bomSize = 2
		} else if len(src) >= 3 &&
			src[0] == utf8BOM[0] &&
			src[1] == utf8BOM[1] &&
			src[2] == utf8BOM[2] {
			d.current = transform.Nop
			bomSize = 3
		}
	}
	if bomSize < len(src) {
		nDst, nSrc, err = d.current.Transform(dst, src[bomSize:], atEOF)
	}
	return nDst, nSrc + bomSize, err
}
```

shueybubbles: ah ha my input file was missing a `N` around the string constant.

shueybubbles: In summary - if it's ok to require a utf16 BOM, we can get away with my change to just use BOMOverride and fallback to the UTF8 decoder.
Otherwise, we would need an additional scan of the file first, or at least a partial scan, to try to detect if it's BOM-less UTF16 file.

shueybubbles: Looking at the ODBC sqlcmd code, I think we can assume UTF16 files have a BOM. Unlike the ODBC one, go-sqlcmd will not use any Windows code page translations as the fallback. 

This is all ODBC sqlcmd does when opening the file:

```C++
HRESULT CTextStream::ReadHeaderEncoding()
{
    BYTE buffer[4] = {0};
    DWORD bytesRead = 0 ;
    m_StartOffset = 0;

    m_FileSize = ::GetFileSize(m_FileHandle, NULL);
    if (m_FileSize != INVALID_FILE_SIZE && m_FileSize >= 4)
    {
        //check for BE,LE,UTF8
        if (!ReadFile(m_FileHandle, buffer, 4, &bytesRead, NULL))
            return HRESULT_FROM_WIN32(GetLastError());

        //little endian unicode file
        if (bytesRead >= 2 && memcmp(buffer, LEStart, 2) == 0)
        {
            //size of header
            m_StartOffset = 2;
            m_CodePage = -1;
        }
        //big endian unicode file
        else if (bytesRead >= 2 && memcmp(buffer, BEStart, 2) == 0)
        {
            //size of header
            m_StartOffset = 2;
            m_bLittleEndian = false;
            m_CodePage = -1;
        }
        //UTF8 format file
        else if (bytesRead >= 3 && memcmp(buffer, UTF8Start, 3) == 0)
        {
            //size of header
            m_StartOffset = 3;
            m_CodePage = CP_UTF8;
        }

        ::SetFilePointer(m_FileHandle, m_StartOffset, NULL, FILE_BEGIN);
    }

    InitFileCache();

    return S_OK;
}
```

shueybubbles:
@simonoff do you mind if I post an alternative PR?

simonoff:
Go a head! I think you better understand it in case of windows encodings:)

shueybubbles:
> Any progress on this one?
> 
> Based on the changes requested, I guess this is only for big endian, what about UTF-16LE then?

@XC- @simonoff can you try out #83 and verify your scenario?

PR #61, opened 2022-03-24
Docker release package

stuartpa:
This PR builds a Docker image for 'sqlcmd'.  To use the docker image simply run (this runs sqlcmd with no args):

  docker run --rm microsoft/sqlcmd

Example of how to pass args:

  docker run --rm microsoft/sqlcmd sqlcmd --help

The PR is designed to work both directly in the local repo and Azure Dev Ops.

This PR is part of a series of 5 initial PRs that will provide packages for all the common managers on the common OSes (Linux, Windows and Mac). Follow up PRs will provide packages for less common configs.

Execute the following command from the root directory of this repository:

./release/linux/docker/pipeline.sh

Output will be sent to ./output/docker

To test the packages:

./release/linux/docker/pipeline-test.sh


build/azure-pipelines/package-common-create.yml
```diff
@@ -32,6 +32,7 @@ steps:
     BUILD_OUTPUT: $(Pipeline.Workspace)
     BUILD_STAGINGDIRECTORY: $(Build.ArtifactStagingDirectory)
 - task: EsrpCodeSigning@1
```
shueybubbles, line 34: code signing is going to vary widely per package type. Perhaps we could move each signing profile into its own yml and include the appropriate yml based on the package type. For a docker package the included yml could have a no-op step.

shueybubbles (approved):

build/azure-pipelines/package-common-create.yml
stuartpa: Agree, I'm about to run into this with the next PR for Windows .msi.

PR #71, opened 2022-04-04
Keep one connection open for the lifetime of sqlcmd

### LINKED ISSUE #68: SQLCMD must stay connected
I started SQLCMD. Then I entered:
```
1> USE Northwind
2> go
Changed database context to 'Northwind'.
1> SELECT * FROM "Order Details"
2> go
Msg 208, Level 16, State 1, Server SOMMERWALD, Line 1
Invalid object name 'Order Details'.
```
And I thought "What?".

Profiler revealed what is going on. For every batch you are connecting and disconnecting. So for the SELECT statement I was back in my default database.

That does not fly. SQLCMD is like SSMS. Once, I have connected, I should stay connected. The above is only one example of what goes wrong. I may create a temp table which I want to use in the next batch. This fails. I may start a transaction that I want to run over multiple batches. But that transaction is rolled back with the first batch. (Which means that only the rest of the transaction is performed, which is really bad.)

(To be precise, SSMS has an option that permits you to disconnect on every query. But that is not a default setting.)

shueybubbles:
Fixes #68

### DIFF
--- a/pkg/sqlcmd/format_test.go
+++ b/pkg/sqlcmd/format_test.go
@@ -4,6 +4,7 @@
 package sqlcmd
 
 import (
+	"context"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


shueybubbles:
I don't know what the right behavior should be if the connection goes bad. I think as-written it would require the user to use `:connect` to reconnect.

stuartpa (approved):

PR #73, opened 2022-04-05
fix -h switch

### LINKED ISSUE #67: sqlcmd -? does not work and -h does the wrong thing.
If I run `sqlcmd -?` with the established SQLCMD I get a list of option. If I run this with go-sqlcmd, I get:

sqlcmd: error: unknown flag -?, did you mean one of "-h", "-c", "-C", "-d", "-E", "-U", "-i", "-o", "-q", "-Q", "-S", "-X", "-G", "-x", "-v", "-a", "-l", "-H", "-K", "-N", "-b", "-V", "-m", "-F", "-r"?

`sqlcmd -h` or `sqlcmd --help` gives the desired result. However, `-h` has an established meaning with SQLCMD to specify how often you want headers printed. I would guess that there quite a few bat files out there with `SQLCMD -h-1` in them. These are people who use SQLCMD to export data and the argument -1 suppresses the headers entirely.

`--help` can still be used for help, but it is not a good idea to use `-h`.

shueybubbles:
Fixes #67 
thx to @stuartpa for blazing the trail

### DIFF
--- a/cmd/sqlcmd/main.go
+++ b/cmd/sqlcmd/main.go
@@ -49,7 +49,7 @@ type SQLCmdArguments struct {
 	ErrorLevel                  int               `short:"m" help:"Controls which error messages are sent to stdout. Messages that have severity level greater than or equal to this level are sent."`
 	Format                      string            `short:"F" help:"Specifies the formatting for results." default:"horiz" enum:"horiz,horizontal,vert,vertical"`
 	ErrorsToStderr              int               `short:"r" help:"Redirects the error message output to the screen (stderr). A value of 0 means messages with severity >= 11 will b redirected. A value of 1 means all error message output including PRINT is redirected." enum:"-1,0,1" default:"-1"`
-	Help                        bool              `short:"?" help:"show syntax summary"`
+	Help                        bool              `short:"?" help:"Show syntax summary."`
 	Headers                     int               `short:"h" help:"Specifies the number of rows to print between the column headings. Use -h-1 to specify that headers not be printed."`
 }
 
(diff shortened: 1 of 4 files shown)


cmd/sqlcmd/main.go
```diff
@@ -49,14 +49,18 @@ type SQLCmdArguments struct {
...
 }
 
 // Validate accounts for settings not described by Kong attributes
 func (a *SQLCmdArguments) Validate() error {
 	if a.PacketSize != 0 && (a.PacketSize < 512 || a.PacketSize > 32767) {
 		return fmt.Errorf(`'-a %d': Packet size has to be a number between 512 and 32767.`, a.PacketSize)
 	}
-
+	if a.Headers < -1 {
+		return fmt.Errorf(`'-h %d': header value must be either -1 or a value between -1 and 2147483647`, a.Headers)
```
stuartpa, line 62: Should this be:

header value must be either -1 or a value between **0** and 2147483647

cmd/sqlcmd/main.go
```diff
@@ -49,14 +49,18 @@ type SQLCmdArguments struct {
 	ErrorLevel                  int               `short:"m" help:"Controls which error messages are sent to stdout. Messages that have severity level greater than or equal to this level are sent."`
 	Format                      string            `short:"F" help:"Specifies the formatting for results." default:"horiz" enum:"horiz,horizontal,vert,vertical"`
 	ErrorsToStderr              int               `short:"r" help:"Redirects the error message output to the screen (stderr). A value of 0 means messages with severity >= 11 will b redirected. A value of 1 means all error message output including PRINT is redirected." enum:"-1,0,1" default:"-1"`
+	Help                        bool              `short:"?" help:"show syntax summary"`
```
stuartpa, line 52: end with '.'?

stuartpa (approved):

dzsquared (approved):

cmd/sqlcmd/main.go
shueybubbles: 0 is invalid too. I copy/pasted the error from the original sqlcmd and it's wrong there
```
e:\>sqlcmd -h 0
Sqlcmd: '-h 0': header value must be either -1 or a value between -1 and 2147483647
```

I'll use 1

Issue #80, opened 2022-04-13
Consider enabling custom date and time format strings provided by the user


shueybubbles:
We don't have a way to get the user's locale information, particularly on non-Windows platforms. 
As part of the fix for #77 I am going to implement the standard format strings for the TSQL date and time types. It'd be cool if users could provide their own format strings mapped to a given SQL type.

For example, the default format string in go for a `DATETIME` will be `2006-01-02 15:04:05.000`

the `time` package in Go defines a bunch of constants for common layouts:

```go
const (
	Layout      = "01/02 03:04:05PM '06 -0700" // The reference time, in numerical order.
	ANSIC       = "Mon Jan _2 15:04:05 2006"
	UnixDate    = "Mon Jan _2 15:04:05 MST 2006"
	RubyDate    = "Mon Jan 02 15:04:05 -0700 2006"
	RFC822      = "02 Jan 06 15:04 MST"
	RFC822Z     = "02 Jan 06 15:04 -0700" // RFC822 with numeric zone
	RFC850      = "Monday, 02-Jan-06 15:04:05 MST"
	RFC1123     = "Mon, 02 Jan 2006 15:04:05 MST"
	RFC1123Z    = "Mon, 02 Jan 2006 15:04:05 -0700" // RFC1123 with numeric zone
	RFC3339     = "2006-01-02T15:04:05Z07:00"
	RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"
	Kitchen     = "3:04PM"
	// Handy time stamps.
	Stamp      = "Jan _2 15:04:05"
	StampMilli = "Jan _2 15:04:05.000"
	StampMicro = "Jan _2 15:04:05.000000"
	StampNano  = "Jan _2 15:04:05.000000000"
)
```

We could define either environment variables or create a new config file for sqlcmd where users can set the format string associated with each TSQL date type. If we use environment variables we can incorporate them into the regular set of SQLCMD variables so scripts could assign it using `:setvar`

PR #82, opened 2022-04-21
Fix rendering of dynamic time values

### LINKED ISSUE #77: The presentation of date/time data types needs some polishing
I would guess this is on a to-do-list, but I file it never the less.

The presentation of datetime values is a little funny as demonstrated by the screenshot below.

- For datetime, +00 is added if the fraction is 0.
- For datetime, no trailing zero is printed when last digit in fraction is 0.
- For date, hours and minutes are printed.
- For datetime2 a +0000 UTC is added that should not be there.
- For datetime2(3) .000 is not printed.
- For time a date of 0000-01-01 is added. Likewise a time zone which should be there is added.

image

shueybubbles:
Fixes #77 
I had to make a driver change first, so now go-sqlcmd will consume `microsoft/go-mssqldb`


pkg/sqlcmd/format.go
```diff
@@ -451,7 +461,26 @@ func (f *sqlCmdFormatterType) scanRow(rows *sql.Rows) ([]string, error) {
 				row[n] = x
 			case time.Time:
 				// Go lacks any way to get the user's preferred time format or even the system default
```
kenvanhyning, line 463: If I am reading this correctly, there is now a way to get the format string, yes?  Can we strike this comment now?

kenvanhyning (approved):

pkg/sqlcmd/format.go
shueybubbles: No, I simply used the same format SQL server uses by default, modified by the precision value for the number of decimal places.
I opened #80 to discuss ways to allow user-provided format strings.

PR #83, opened 2022-05-02
Support UTF 16 input files

shueybubbles:
I modified the proposal in #60 a little bit and added a test.


pkg/sqlcmd/sqlcmd.go
```diff
@@ -32,6 +34,8 @@ var (
 	ErrCtrlC = errors.New(WarningPrefix + "The last operation was terminated because the user pressed CTRL+C")
 )
 
+const maxLineBuffer = 2 * 1024 * 1024 // 2Mb
```
shueybubbles, line 37: Should this be a default value that can be overridden by a property on sqlcmd? Is there a potential performance benefit for apps that know they are using smaller statements and want to set the value smaller?

dzsquared: if we're seeing go-sqlcmd use in IoT environments with really small compute modules, I could see this being true - but overall I think a 2Mb buffer is a reasonable setting to keep constant

dzsquared (approved):

stuartpa (approved):

Issue #87, opened 2022-05-16
shell behaves erratically after exiting go-sqlcmd


kiilivi:
I have tried to use go-sqlcmd in Macbook Pro M1 Docker, but it doesn't seem to be working. My Dockerfile is based on Fedora 36. This is how I installed go-sqlcmd:
```
RUN curl -L https://github.com/microsoft/go-sqlcmd/releases/download/v0.7.0/sqlcmd-v0.7.0-linux-arm64.tar.bz2 > sqlcmd-v0.7.0-linux-arm64.tar.bz2
RUN tar jxf sqlcmd-v0.7.0-linux-arm64.tar.bz2 -C /usr/bin
RUN rm sqlcmd-v0.7.0-linux-arm64.tar.bz2
```

If I run the sqlcmd command without parameters in Docker container, it exits with the following error:
```
unable to open tcp connection with host 'localhost:1433': dial tcp 127.0.0.1:1433: connect: connection refused
```
That's ok, but somehow the shell behaves erraneously after that.

If I run the sqlcmd with proper parameters, it doesn't seem to do anything. No errors, no output, but it certainly doesn't update database either. I use this kind of command:
```
/usr/bin/sqlcmd -S host.docker.internal -U <user> -d <database> -i <script>.sql  2>&1
```
I use the same command line with MacOS sqlcmd (original version, not go-sqlcmd) and it works just fine.

Is this a bug or I'm doing something wrong?

stuartpa:
Is it possible you SSL/TLS is not configured for the target server to allow an encrypted connection, if so does adding the switch "-N disable" help? 

This error is the error I'd expect in the case an encrypted connection cannot be established

```
unable to open tcp connection with host 'localhost:1433': dial tcp 127.0.0.1:1433: connect: connection refused
```

This implementation of sqlcmd has "encrypt connection" set to the default (per breaking changes, and the -?)

kiilivi:
I was a bit unclear... I wanted to show that error message in order to prove that I'm using the right binary. The database is running outside the docker container, so there's no service at 127.0.0.1 1433, so there's nothing wrong with the error message. 

I managed to sort out most of my problems (I was using go-sqlcmd in a Robot Framework script and the error was a bit concealed), but go-sqlcmd still makes bash shell command line behave erratically after exiting sqlcmd. I can live with that, since I run it from scripts, but it is weird behaviour.

stuartpa:
[...]

Issue #39, opened 2021-12-10
Input file option (-i) is being ignored


arvindshmicrosoft:
I am trying to use go-sqlcmd to execute a script (.SQL file). The command line I am using is below:

```
sqlcmd.exe -i "C:\somedir\somescript.sql" --authentication-method=ActiveDirectoryManagedIdentity -S someserver.database.windows.net
```

For some reason the sqlcmd exits silently without throwing any errors. I tried specifying non-existent file paths and the result is the same. I suspect there is some issue with the way sqlcmd is ignoring the -i parameter. this is go-sqlcmd built from main (commit hash 32462b7a4e59f35479550702032763ba1a66905a).

I don't see any issues with `-Q` with the same authentication. Something seems specifically wrong with `-i`

shueybubbles:
Try v0.8.0

PR #88, opened 2022-05-20
exit on state 127

### LINKED ISSUE #76: SQLCMD should exit on certain errors
Consider:

RAISERROR('This is the end', 1, 127)

When you run this in the old SQLCMD, the scripts exits, because state is 127. go-sqlcmd does not. (This an old holdover from ISQL, and should still work I think. No, this does not work in SSMS.)

On a similar note:

RAISERROR('Get me out of here!', 20, 1) WITH LOG

This closes the connection to SQL Server, and old SQLCMD exits at this point. go-sqlcmd does not, but nor does it reconnect, so you only get more error messages.

shueybubbles:
Fixes #76 
Had to pick up new `sqlexp` and `go-mssqldb`


shueybubbles:
also fixes #87 (I hope)

shueybubbles:
Fixes #39

pkg/sqlcmd/sqlcmd.go
```diff
@@ -44,6 +44,8 @@ type Console interface {
 	ReadPassword(prompt string) ([]byte, error)
 	// SetPrompt sets the prompt text shown to input the next line
 	SetPrompt(s string)
+	// Close clears any buffers and closes open file handles
+	Close()
```
stuartpa, line 48: I'm not seeing an implementation for this method, is this by design?

stuartpa (approved):

pkg/sqlcmd/sqlcmd.go
shueybubbles: it's already implemented. https://github.com/microsoft/go-sqlcmd/blob/e3972914801bc3492b49b6fe642a7e931e088be3/pkg/console/console.go#L39

Issue #94, opened 2022-05-24
Add `--version` flag to CLI


dhirschfeld:
Having a  `--version` flag is pretty standard for a CLI as it lets the user know what version of the tool they're using, and hence also helps with bug reports.
```
> sqlcmd --version
sqlcmd.exe: error: unknown flag --version
```

dhirschfeld:
Am I missing something obvious and there's another way to know what version of `sqlcmd` I'm running?

stuartpa:
Fixed in #100

PR #100, opened 2022-06-22
Add --version flag

stuartpa:
Resolves issue #94


stuartpa:
Right now when built in the pipeline this PR is product a version number like:

sqlcmd --version
sqlcmd: v20220622.2

which is based off the $(Build.BuildNumber) value.

We need it to be:

sqlcmd --version
sqlcmd: v0.8.0-1

I'm investigating

build/azure-pipelines/build-common.yml
```diff
@@ -45,7 +45,7 @@ steps:
   displayName: 'Go: build sqlcmd'
   inputs:
     command: 'build'
-    arguments: '-o $(Build.BinariesDirectory)'
+    arguments: '-o $(Build.BinariesDirectory) -ldflags="-X main.version=$(Build.BuildNumber)"'
```
shueybubbles, line 48: Look at build\azure-pipelines\build-product.yml

```yml
      steps:
      - task: PowerShell@2
        displayName: Set last tag to variable
        inputs:
          targetType: 'inline'
          script: |
            $VERSION_TAG = git describe --tags (git rev-list --tags --max-count=1)
            Write-Host("##vso[task.setvariable variable=VERSION_TAG]$VERSION_TAG")
            Write-Host($VERSION_TAG)
```

This should also set the build number for the pipeline for the build to pickup

shueybubbles:
Can you fix the diff not to have so much whitespace change?

build/azure-pipelines/build-common.yml
```diff
@@ -8,7 +8,16 @@ parameters:
 - name: ArtifactName
   type: string
 
-steps: 
+steps:
+- task: PowerShell@2
```
shueybubbles, line 12: instead of duplicating what's already in build-product.yml, make this value a parameter on the template and have build-product.yml pass in the value.

shueybubbles: the default value for the parameter can be $(Build.BuildNumber) just in case some pipeline that uses this yml doesn't set it. We only care about the tag for official builds

shueybubbles (changes requested):
:clock1:

build/azure-pipelines/build-common.yml
stuartpa: Ack

build/azure-pipelines/build-common.yml
```diff
@@ -7,8 +7,11 @@ parameters:
   default: 
 - name: ArtifactName
   type: string
+- name: VersionTag
+  type: string
```
shueybubbles, line 11: you're missing the update to build-product.yml to pass the value

stuartpa: I missed a git add!

shueybubbles:
>         displayName: Set last tag to variable

i think this task has to move into the first stage

---
Refers to: build/azure-pipelines/build-product.yml:58 in 75c7dfe. [](commit_id = 75c7dfedf813e21607454c8e5a330b113d9b5c0f, deletion_comment = False)

shueybubbles:
>       - v*

add a branch constraint here for main

---
Refers to: build/azure-pipelines/build-product.yml:4 in 75c7dfe. [](commit_id = 75c7dfedf813e21607454c8e5a330b113d9b5c0f, deletion_comment = False)

stuartpa:
> > ```
> >   - v*
> > ```
> 
> add a branch constraint here for main unless we are confident we won't push version tags to other branches incorrectly.
> 
> Refers to: build/azure-pipelines/build-product.yml:4 in [75c7dfe](https://github.com/microsoft/go-sqlcmd/commit/75c7dfedf813e21607454c8e5a330b113d9b5c0f). [](commit_id = [75c7dfe](https://github.com/microsoft/go-sqlcmd/commit/75c7dfedf813e21607454c8e5a330b113d9b5c0f), deletion_comment = False)

Did you add this yesterday as part of another PR?

PR #121, opened 2022-08-19
Initialize console if -U is specified

### LINKED ISSUE #115: Password prompt is not displayed when -i and -U switch are specified
Password prompt is not shown when used with -i and -U switch

Windows
```
>.\go-sqlcmd\sqlcmd.exe -i go-sqlcmd\input.sql -U sa
mssql: login error: Login failed for user 'sa'.
```
Linux
```
$ ./sqlcmd -S 172.20.112.1 -U sa -i testvar.sql
mssql: login error: Login failed for user 'sa'.
```

Behavior with ODBC sqlcmd
```
C:\Users\apdeshmukh\git>sqlcmd -i go-sqlcmd\input.sql -U sa
Password:
```

apoorvdeshmukh:
In absence of console or when console reference line is set
to nil, noPwd is set to true which disables password prompt
and when console is not initialized, password cannot be read
from user. This is why when -i is specified, console is not
initialized and we do not see password prompt.
This commit initializes console even if -U is specified so
that password can be read interactively from user.
Resolves #115 

Validation:
```
go-sqlcmd>.\sqlcmd.exe -i test.sql -U sa
Password:



------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Microsoft SQL Server 2012 (SP4-GDR) (KB4583465) - 11.0.7507.2 (X64)
        Nov  1 2020 00:48:37
        Copyright (c) Microsoft Corporation
        Enterprise Edition: Core-based Licensing (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)


(1 row affected)
```

### DIFF
--- a/cmd/sqlcmd/main.go
+++ b/cmd/sqlcmd/main.go
@@ -211,17 +211,17 @@ func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 		return 1, err
 	}
 
-	iactive := args.InputFile == nil && args.Query == ""
-	uactive := args.UserName != ""
+	var connectConfig sqlcmd.ConnectSettings
+	setConnect(&connectConfig, args, vars)
 	var line sqlcmd.Console = nil
-	if iactive || uactive {
+	if connectConfig.RequiresPassword() {
 		line = consoleAllocator("")
 		defer line.Close()
 	}
 
 	s := sqlcmd.New(line, wd, vars)
 	s.UnicodeOutputFile = args.UnicodeOutputFile
-	setConnect(&s.Connect, args, vars)
+
 	if args.BatchTerminator != "GO" {
 		err = s.Cmd.SetBatchTerminator(args.BatchTerminator)
 		if err != nil {
@@ -266,6 +266,9 @@ func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 	if err != nil {
 		return 1, err
 	}
+
+	iactive := args.InputFile == nil && args.Query == ""
+
 	if iactive || s.Query != "" {
 		err = s.Run(once, false)
 	} else {
--- a/cmd/sqlcmd/main.go
+++ b/cmd/sqlcmd/main.go
@@ -201,9 +201,10 @@ func setConnect(connect *sqlcmd.ConnectSettings, args *SQLCmdArguments, vars *sq
 	connect.ErrorSeverityLevel = args.ErrorSeverityLevel
 }
 
-type allocConsole func(historyFile string) sqlcmd.Console
-
-var consoleAllocator allocConsole = console.NewConsole
+func IsConsoleInitializationRequired(connect *sqlcmd.ConnectSettings, args *SQLCmdArguments) bool {
+	iactive := args.InputFile == nil && args.Query == ""
+	return iactive || connect.RequiresPassword()
+}
 
 func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 	wd, err := os.Getwd()
@@ -214,8 +215,8 @@ func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 	var connectConfig sqlcmd.ConnectSettings
 	setConnect(&connectConfig, args, vars)
 	var line sqlcmd.Console = nil
-	if connectConfig.RequiresPassword() {
-		line = consoleAllocator("")
+	if IsConsoleInitializationRequired(&connectConfig, args) {
+		line = console.NewConsole("")
 		defer line.Close()
 	}
 
@@ -232,7 +233,7 @@ func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 		return 1, err
 	}
 
-	setConnect(&s.Connect, args, vars)
+	s.Connect = &connectConfig
 	s.Format = sqlcmd.NewSQLCmdDefaultFormatter(false)
 	if args.OutputFile != "" {
 		err = s.RunCommand(s.Cmd["OUT"], []string{args.OutputFile})
@@ -268,7 +269,6 @@ func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 	}
 
 	iactive := args.InputFile == nil && args.Query == ""
-
 	if iactive || s.Query != "" {
 		err = s.Run(once, false)
 	} else {
--- a/cmd/sqlcmd/main.go
+++ b/cmd/sqlcmd/main.go
@@ -208,8 +208,9 @@ func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 	}
 
 	iactive := args.InputFile == nil && args.Query == ""
+	uactive := args.UserName != ""
 	var line sqlcmd.Console = nil
-	if iactive {
+	if iactive || uactive {
 		line = console.NewConsole("")
 		defer line.Close()
 	}
@@ -257,7 +258,7 @@ func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 		s.Query = args.Query
 	}
 	// connect using no overrides
-	err = s.ConnectDb(nil, !iactive)
+	err = s.ConnectDb(nil, line == nil)
 	if err != nil {
 		return 1, err
 	}
--- a/cmd/sqlcmd/main.go
+++ b/cmd/sqlcmd/main.go
@@ -201,6 +201,10 @@ func setConnect(connect *sqlcmd.ConnectSettings, args *SQLCmdArguments, vars *sq
 	connect.ErrorSeverityLevel = args.ErrorSeverityLevel
 }
 
+type allocConsole func(historyFile string) sqlcmd.Console
+
+var consoleAllocator allocConsole = console.NewConsole
+
 func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 	wd, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


shueybubbles:
pls add a test that fails without the fix.

cmd/sqlcmd/main.go
```diff
@@ -102,6 +102,17 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
 	return a.AuthenticationMethod
 }
 
+func getSqlcmdConsole(args SQLCmdArguments) sqlcmd.Console {
+	iactive := args.InputFile == nil && args.Query == ""
+	uactive := args.UserName != ""
+	var sqlcmdConsole sqlcmd.Console = nil
+	if iactive || uactive {
+		sqlcmdConsole = console.NewConsole("")
+		defer sqlcmdConsole.Close()
```
shueybubbles, line 111: this will close the console as soon as getSqlcmdConsole returns

cmd/sqlcmd/main_test.go
```diff
@@ -162,8 +162,8 @@ func TestUnicodeOutput(t *testing.T) {
 	}
 	vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
 	setVars(vars, &args)
-
-	exitCode, err := run(vars, &args)
+	sqlcmdConsole := getSqlcmdConsole(args)
```
shueybubbles, line 165: let's go with a model that doesn't require so many changes.
Just define a type like
```go

type allocConsole func(historyFile string) sqlcmd.Console

```
And create a field whose default assignment is console.NewConsole. The test can override it by assigning it a new func.

shueybubbles (changes requested):

cmd/sqlcmd/main.go
```diff
@@ -111,6 +117,12 @@ func main() {
 	vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
 	setVars(vars, &args)
 
+	initializeConsole := (args.InputFile == nil && args.Query == "") || args.UserName != ""
+	if initializeConsole {
+		setConsole(console.NewConsole(""))
+		defer line.Close()
+	}
```
shueybubbles, line 124: why does this move out of run()? the tests call run() directly.

apoorvdeshmukh: I'm reworking the changes. Will be making `line` a global variable and it will be set via a function.

shueybubbles: the global variable should probably be a func type.

cmd/sqlcmd/main_test.go
```diff
@@ -327,6 +328,64 @@ func TestMissingInputFile(t *testing.T) {
...
+	return tc.OnPasswordPrompt(prompt)
+}
+
+func (tc *testConsole) SetPrompt(s string) {
+	tc.PromptText = s
+}
+
+func (tc *testConsole) Close() {
+}
+
```
shueybubbles, line 388: can we share a testConsole implementation with the sqlcmd_test? Perhaps by creating a new package under <root>/internal then have both tests include that package.

cmd/sqlcmd/main.go
```diff
@@ -201,16 +201,21 @@ func setConnect(connect *sqlcmd.ConnectSettings, args *SQLCmdArguments, vars *sq
...
+var consoleAllocator allocConsole = console.NewConsole
+
 func run(vars *sqlcmd.Variables, args *SQLCmdArguments) (int, error) {
 	wd, err := os.Getwd()
 	if err != nil {
 		return 1, err
 	}
 
 	iactive := args.InputFile == nil && args.Query == ""
+	uactive := args.UserName != ""
```
shueybubbles, line 215: >.UserName != "" [](http://example.com/codeflow?start=16&length=15)

i think username can be non-empty and password be empty for some versions of AAD auth.

cmd/sqlcmd/main.go
```diff
@@ -201,22 +201,27 @@ func setConnect(connect *sqlcmd.ConnectSettings, args *SQLCmdArguments, vars *sq
...
 		return 1, err
 	}
 
-	iactive := args.InputFile == nil && args.Query == ""
+	var connectConfig sqlcmd.ConnectSettings
+	setConnect(&connectConfig, args, vars)
 	var line sqlcmd.Console = nil
-	if iactive {
-		line = console.NewConsole("")
+	if connectConfig.RequiresPassword() {
```
shueybubbles, line 217: don't you still need the iactive check? How are you testing this before pushing?

apoorvdeshmukh: My bad. I missed that locally it is only using SQL authentication.
Adding testcases for AAD modes

shueybubbles:
> 	setConnect(&s.Connect, args, vars)

i think this can change to copy connectConfig to s.Connect instead

---
Refers to: cmd/sqlcmd/main.go:235 in 99140d4. [](commit_id = 99140d449fc38690fd41e74b9cfe4ece24966c18, deletion_comment = False)

cmd/sqlcmd/main_test.go
apoorvdeshmukh: The unit test now no more relies on the console allocator function hence have removed it.
The unit test will validate console initialization and conditions for password prompt based on the parameters supplied.

cmd/sqlcmd/main.go
apoorvdeshmukh: Added the iactive check.

cmd/sqlcmd/main.go
```diff
@@ -201,22 +201,28 @@ func setConnect(connect *sqlcmd.ConnectSettings, args *SQLCmdArguments, vars *sq
 	connect.ErrorSeverityLevel = args.ErrorSeverityLevel
 }
 
+func IsConsoleInitializationRequired(connect *sqlcmd.ConnectSettings, args *SQLCmdArguments) bool {
```
shueybubbles, line 204: >IsCons [](http://example.com/codeflow?start=5&length=6)

isConsoleInitializationRequired
no need for this to be public

cmd/sqlcmd/main_test.go
```diff
@@ -327,6 +327,27 @@ func TestMissingInputFile(t *testing.T) {
...
 
+func TestConditionsForPasswordPrompt(t *testing.T) {
+
+	args := newArguments()
+	args.InputFile = []string{"testdata/someFile.sql"}
+	args.DisableCmdAndWarn = true
+	vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
+	setVars(vars, &args)
+	var connectConfig sqlcmd.ConnectSettings
+	setConnect(&connectConfig, &args, vars)
```
shueybubbles, line 338: Let's not use setVars and setConnect in this test. Their behavior depends on the values of environment variables set in the pipeline.
Just set the specific combinations of authentication type and user name/password in each struct explicitly and validate the right values for isConsoleInitializationRequired.

pkg/sqlcmd/commands_test.go
```diff
@@ -167,7 +167,7 @@ func TestConnectCommand(t *testing.T) {
 	err := connectCommand(s, []string{"someserver -U someuser"}, 1)
 	assert.NoError(t, err, "connectCommand with valid arguments doesn't return an error on connect failure")
 	assert.True(t, prompted, "connectCommand with user name and no password should prompt for password")
-	assert.NotEqual(t, "someserver", s.Connect.ServerName, "On error, sqlCmd.Connect does not copy inputs")
+	assert.Equal(t, "someserver", s.Connect.ServerName, "servername should match with the input parameter")
```
shueybubbles, line 171: why is this changing?

cmd/sqlcmd/main_test.go
apoorvdeshmukh: setConnect only seems to set password from env variable and that too if `DisableCmdAndWarn` is false.
servername and usernamre is fetched from vars.get() vars is obtained from InitializeVariables() which has an argument to indicate if values should be fetched from end. In this case DisableCmdAndWarn is set to true and value of DisableCmdAndWarn is passed to ensure that we don't fetch anything from env.
setVars() doesn't seems to fetch any anything from env.
I will, however, parameterize the test to add more scenarios.

pkg/sqlcmd/commands_test.go
apoorvdeshmukh: Thanks for adding the context. This was a side effect of changing type of connect from ConnectSettings to a pointer to ConnectSettings in sqlcmd struct.
In connectCommand() we copy s.Connect to local connect object to establish connection.
I will fix this in my next commit.

shueybubbles: if the message in the assert didn't provide sufficient context, please update the assert message to be clearer.

cmd/sqlcmd/main_test.go
```diff
@@ -327,6 +328,54 @@ func TestMissingInputFile(t *testing.T) {
...
+		args.InputFile = testcase.inputFile
+		args.UserName = testcase.username
+		vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
+		setVars(vars, &args)
+		var connectConfig sqlcmd.ConnectSettings
+		setConnect(&connectConfig, &args, vars)
+		connectConfig.AuthenticationMethod = testcase.authenticationMethod
+		connectConfig.Password = testcase.pwd
+		assert.Equal(t, testcase.expectedResult, isConsoleInitializationRequired(&connectConfig, &args), "Unexpected test result encountered for console initialization")
+		assert.Equal(t, testcase.expectedResult, connectConfig.RequiresPassword() && connectConfig.Password == "", "Unexpected test result encountered for password prompt conditions")
```
shueybubbles, line 375: >sse [](http://example.com/codeflow?start=3&length=3)

this assert is odd. Shouldn't you assert that if RequiresPassword is true then Password == ""?

shueybubbles (approved):
:shipit:

cmd/sqlcmd/main_test.go
apoorvdeshmukh: Yes. This was to avoid the conditional check and to assert the same condition for password prompt that we use in connectDb()

PR #122, opened 2022-08-19
Distinguish non-error output from error output

### LINKED ISSUE #105: Errors for :R are thrown on stdout instead of stderr
When compared with the ODBC based sqlcmd utility, the go-sqlcmd utility throws error for :R and all other commands to stdout instead of stderr.
This issue is created to align go-sqlcmd behavior with ODBC based sqlcmd.

Usage on ODBC sqlcmd
```
>type test1.sql
:r missing.sql
go

>sqlcmd -i test1.sql >stdout.txt 2>stderr.txt
stdout: <Blank>
stderr: 'missing.sql': Invalid filename.
```

Usage on go-sqlcmd
```
sqlcmd-v0.8.0-windows-x64>sqlcmd.exe -i test.sql >stdout.txt 2>stderr.txt
stdout: open missing.sql: The system cannot find the file specified.
stderr: <blank>
```

apoorvdeshmukh:
In certain cases such as non-interactive mode it was observed
that error messages meant for error stream were captured under
output stream. This is not desirable if the user only wants to
retrieve the error messages. This commit aligns the go-sqlcmd
behaviour with ODBC sqlcmd where error messages are captured
in the appropriate error stream.
Resolves #105 

Validation:
```
go-sqlcmd>type test.sql
:r missing.sql
go
C:\Users\apdeshmukh\git\go-sqlcmd>type test.sql
:r missing.sql
go
select @@version
go
go-sqlcmd>.\sqlcmd.exe -i test.sql >stdout.txt 2>stderr.txt

go-sqlcmd>type stdout.txt


------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Microsoft SQL Server 2019 (RTM-GDR) (KB5014356) - 15.0.2095.3 (X64)
        Apr 29 2022 18:00:13
        Copyright (C) 2019 Microsoft Corporation
        Developer Edition (64-bit) on Windows 10 Enterprise 10.0 <X64> (Build 22000: ) (Hypervisor)


(1 row affected)

go-sqlcmd>type stderr.txt
Sqlcmd: Error:  Error occurred while opening or operating on file missing.sql (Reason: open missing.sql: The system cannot find the file specified.).

```

### DIFF
--- a/pkg/sqlcmd/sqlcmd.go
+++ b/pkg/sqlcmd/sqlcmd.go
@@ -85,6 +85,8 @@ func New(l Console, workingDirectory string, vars *Variables) *Sqlcmd {
 	s.PrintError = func(msg string, severity uint8) bool {
 		return false
 	}
+	s.SetOutput(os.Stdout)
+	s.SetError(os.Stderr)
 	return s
 }
 
@@ -133,7 +135,11 @@ func (s *Sqlcmd) Run(once bool, processAll bool) error {
 				args = make([]string, 0)
 				once = true
 			} else {
-				_, _ = s.GetOutput().Write([]byte(err.Error() + SqlcmdEol))
+				if iactive {
+					_, _ = s.GetOutput().Write([]byte(err.Error() + SqlcmdEol))
+				} else {
+					_, _ = s.GetError().Write([]byte(err.Error() + SqlcmdEol))
+				}
 			}
 		}
 		if cmd != nil {
@@ -143,7 +149,11 @@ func (s *Sqlcmd) Run(once bool, processAll bool) error {
 				break
 			}
 			if err != nil {
-				_, _ = s.GetOutput().Write([]byte(err.Error() + SqlcmdEol))
+				if iactive {
+					_, _ = s.GetOutput().Write([]byte(err.Error() + SqlcmdEol))
+				} else {
+					_, _ = s.GetError().Write([]byte(err.Error() + SqlcmdEol))
+				}
 				lastError = err
 			}
 		}
--- a/pkg/sqlcmd/sqlcmd_test.go
+++ b/pkg/sqlcmd/sqlcmd_test.go
@@ -109,6 +109,28 @@ func TestSqlCmdQueryAndExit(t *testing.T) {
 	}
 }
 
+func TestSqlCmdOutputAndError(t *testing.T) {
+	s, outfile, errfile := setupSqlcmdWithFileErrorOutput(t)
+	defer os.Remove(outfile.Name())
+	defer os.Remove(errfile.Name())
+	s.Query = "select $(X"
+	err := s.Run(true, false)
+	if assert.NoError(t, err, "s.Run(once = true)") {
+		bytes, err := os.ReadFile(errfile.Name())
+		if assert.NoError(t, err, "os.ReadFile") {
+			assert.Equal(t, "Sqlcmd: Error: Syntax error at line 1."+SqlcmdEol, string(bytes), "Incorrect output from Run")
+		}
+	}
+	s.Query = "select '1'"
+	err = s.Run(true, false)
+	if assert.NoError(t, err, "s.Run(once = true)") {
+		bytes, err := os.ReadFile(outfile.Name())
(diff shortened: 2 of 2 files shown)


pkg/sqlcmd/sqlcmd.go
```diff
@@ -133,7 +135,11 @@ func (s *Sqlcmd) Run(once bool, processAll bool) error {
 				args = make([]string, 0)
 				once = true
 			} else {
-				_, _ = s.GetOutput().Write([]byte(err.Error() + SqlcmdEol))
+				if iactive {
+					_, _ = s.GetOutput().Write([]byte(err.Error() + SqlcmdEol))
+				} else {
+					_, _ = s.GetError().Write([]byte(err.Error() + SqlcmdEol))
```
shueybubbles, line 141: Don't you need to distinguish the case of iactive being true because of the user name parameter being passed in, which is part of your other PR?

shueybubbles: maybe Sqlcmd should have a property like "WriteErrorsToErrorStream bool" that would be set by main before it calls Run.

There are some other settings that control command output destination that this could be like.

apoorvdeshmukh: Added flag in sqlcmd struct to indicate an interactive session.
The interactive session will be determined when the first input is scanned.

pkg/sqlcmd/sqlcmd_test.go
```diff
@@ -109,6 +109,28 @@ func TestSqlCmdQueryAndExit(t *testing.T) {
 	}
 }
 
+func TestSqlCmdOutputAndError(t *testing.T) {
+	s, outfile, errfile := setupSqlcmdWithFileErrorOutput(t)
```
shueybubbles, line 113: don't you need to test the s.IncludeFile path too?

apoorvdeshmukh: Added test for s.IncludeFile

pkg/sqlcmd/sqlcmd.go
```diff
@@ -86,10 +87,13 @@ func New(l Console, workingDirectory string, vars *Variables) *Sqlcmd {
 	s.PrintError = func(msg string, severity uint8) bool {
 		return false
 	}
+	s.SetOutput(os.Stdout)
+	s.SetError(os.Stderr)
 	return s
 }
 
 func (s *Sqlcmd) scanNext() (string, error) {
+	s.IsInteractiveSession = true
```
shueybubbles, line 96: add a comment on why this is being set

pkg/sqlcmd/sqlcmd.go
```diff
@@ -86,10 +87,13 @@ func New(l Console, workingDirectory string, vars *Variables) *Sqlcmd {
 	s.PrintError = func(msg string, severity uint8) bool {
 		return false
 	}
+	s.SetOutput(os.Stdout)
+	s.SetError(os.Stderr)
```
shueybubbles, line 91: >s.SetError(os.Stderr) [](http://example.com/codeflow?start=1&length=21)

this is a change in behavior for any other places that call GetError(), now it's going to return stderr by default instead of stdout. Please verify that's the right behavior in those other places.

pkg/sqlcmd/sqlcmd.go
shueybubbles: would it make sense to fix s.GetError to do the right thing based on context and just change this code to use GetError() only? 
I'm not 100% certain that every kind of error in sqlcmd goes to the same stream, though.

shueybubbles (approved):
:shipit:

pkg/sqlcmd/sqlcmd.go
apoorvdeshmukh: This functionality change is leading to undesired side effects such as redirecting server errors to stderr as well. (I will close this PR and create new one)
I have reworked the changes to keep existing functionality as is in PR #143 and redirecting only the sqlcmd errors which is what the ODBC is also doing.

PR #138, opened 2022-09-20
Modern POSIX style CLI for SQLCMD - Part 2

stuartpa:
This is the 2nd part for the modern-CLI.  See PR165 for first part:

- https://github.com/microsoft/go-sqlcmd/pull/165

This PR moves the file-layout for the modern CLI to conform to the idiomatic golang layout, and focuses on test and code coverage, there is no new functionality in this PR.  The Unit test coverage for the modern sqlcmd.exe now 95%+.

The formatting of the MIT license file was part of @shueybubbles original commit, I presume we want to keep these.

TODO:

- A follow up PR will at the ./cmd/modern folder (that builds the new sqlcmd.exe) to the build pipelines. 
- Right now, three tests are skipped, TestQuery, TestNegUninstallNoInstanceToUninstall and TestUnicodeInput, I am debugging these now


go.mod
```diff
@@ -4,28 +4,58 @@ go 1.18
 
 require (
 	github.com/alecthomas/kong v0.5.1-0.20220516223738-0aaa4c11997b
+	github.com/billgraziano/dpapi v0.4.0
```
shueybubbles, line 7: >/billgraziano/dpapi v0.4.0 [](http://example.com/codeflow?start=11&length=26)

are we going to dump dpapi and just stored clear text?

cmd/bulkcopy.go
```diff
@@ -0,0 +1,60 @@
...
+	SchemaObjectName     string
+	QueryText            string
+	BatchSize            int
+	ErrorFile            string
+	FormatFile           string
+	FirstRow             int64
+	Hints                string
+	InputFile            string
+	LastRow              int64
+	MaxErrors            int
```
shueybubbles, line 23: FYI i'm redoing this struct and putting it in /pkg/bulkcopy

cmd/use-context.go
```diff
@@ -0,0 +1,107 @@
...
+	const short = "Set the current-context in a sqlconfig file."
+	const long = short
+	const example = `# Use the context for the sa@sql1 sql instance
+  sqlcmd config use-context sa@sql1`
+
+	var run = func(cmd *cobra.Command, args []string) {
+		var config Sqlconfig
+		var name = args[0]
+
+		viper.Unmarshal(&config)
```
shueybubbles, line 25: viper has its own methods to search config with xpath-like expressions. I am curious whether those methods are faster than this model of using Unmarshal followed by a bunch of loops in the helper methods.
Or are they so complex to use that any perf benefit they haven't isn't worth the developer effort?

go.mod
stuartpa: Going to encrypted bits to disk, then provide `sqlcmd config view --raw` to view the secret decrypted

cmd/install.go
```diff
@@ -0,0 +1,255 @@
...
+
+func main() {
+}
+
+func generatePassword(passwordLength, minSpecialChar, minNum, minUpperCase int) string {
+	var password strings.Builder
+
+	//Set special character
+	for i := 0; i < minSpecialChar; i++ {
+		random := rand.Intn(len(specialCharSet))
```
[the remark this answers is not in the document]
stuartpa: Resolved, moved to crypto/rand

cmd/bulkcopy.go
stuartpa: Sounds good

cmd/install.go
```diff
@@ -0,0 +1,391 @@
...
+	"strings"
+	"time"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/billgraziano/dpapi"
```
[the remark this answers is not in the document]
stuartpa: Fixed, builds on Linux now.

.gitignore
```diff
@@ -1,5 +1,4 @@
 # Binaries for programs and plugins
-output
```
stuartpa, line 2: Not sure right now if ignoring "output" is needed now.  There is a helpers/output package in this PR which breaks unless we make this ignore more specific, or rename the helpers/output package

build/azure-pipelines/build-common.yml
```diff
@@ -1,67 +1,66 @@
-parameters:
```
stuartpa, line 1: I've just discovered what this diff issue is.  I turned off all auto code cleanup features in my IDE (goland), and there has been no issues.  Then I just ran gofmt -w . and pushed a commit and that has caused the problem.

So I think we need to get through this at some point, as all files should be gofmt'ed.  I can undo this local commit if needed, and then create a seperate PR that gofmt's the code, we should then add a pipeline break if gofmt . fails.

shueybubbles:
> 

we should commit .editorconfig so everyone gets the same behavior. the go tooling handles most of it ok but if there are options the go tools don't enforce automatically we should use this file for consistency.

---
Refers to: .gitignore:27 in 3db8294. [](commit_id = 3db829431465689d481d049b4b091272c20168f3, deletion_comment = False)

build/azure-pipelines/build-common.yml
shueybubbles: the Go extension for vscode always runs fmt on save

cmd/cmd.go
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package cmd
+
+import (
+	. "github.com/microsoft/go-sqlcmd/cmd/commander"
+	"github.com/microsoft/go-sqlcmd/cmd/helpers"
+	"github.com/microsoft/go-sqlcmd/cmd/helpers/output"
```
shueybubbles, line 9: should helpers move from /cmd to /internal?
Doing so would hide public types from other apps that reference this repo. Let's be deliberate about what goes public and what does not. 

I think it's certainly desirable for third parties to reference /pkg/sqlcmd, for example.

cmd/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		output.Info()
+	}
+}
+
+func IsValidRootCommand(command string) (valid bool) {
+	for _, c := range root.Commands {
+		if command == c.GetCommand().Name() {
+			valid = true
+			break
```
shueybubbles, line 78: >break [](http://example.com/codeflow?start=3&length=5)

why not return?
There's no need to run the second loop now.

cmd/defaults_bcp.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+//go:build bcp
+
+package cmd
+
+const appName = "bcp"
+const defaultCommand = "bulkcopy"
```
shueybubbles, line 9: these defaults_xxx files can go away

cmd/helpers/doc.go
```diff
@@ -0,0 +1,53 @@
...
+    trace method to output low level debugging information
+ - Do not return error if client is not going use the error for control flow, call the
+   injected checkErr instead, which will probably end up calling cobra.checkErr and exit:
+     e.g. Do not sprinkle application (non-helper) code with:
+       err, _ := fmt.printf("Hope this works")
+       if (err != nil) {
+         panic("How unlikely")
+       }
+     Do use the injected checkErr callback and let the application decide what to do
+       err, _ := printf("Hope this works)
```
shueybubbles, line 27: > err, _ := printf("Hope this works) [](http://example.com/codeflow?start=6&length=35)

is there a class of functions that return error where it's ok to ignore the return completely? 
So far for pkg/sqlcmd and cmd/sqlcmd I've just ignored all errors from stdout etc.

cmd/helpers/config/config.go
```diff
@@ -0,0 +1,97 @@
...
+
+	contextName = FindUniqueContextName(contextName, username)
+	endPointName := FindUniqueEndpointName(contextName)
+	userName := username + "@" + contextName
+
+	config.ApiVersion = "v1"
+	config.Kind = "Config"
+	config.CurrentContext = contextName
+
+	config.Endpoints = append(config.Endpoints, Endpoint{
```
shueybubbles, line 44: >append [](http://example.com/codeflow?start=20&length=6)

this says "append" but the function name is "Update".
Shouldn't the function also be Append?

cmd/helpers/config/config.go
```diff
@@ -0,0 +1,97 @@
...
+
+package config
+
+import (
+	. "github.com/microsoft/go-sqlcmd/cmd/sqlconfig"
+)
+
+var config Sqlconfig
+
+func Update(
```
shueybubbles, line 12: >unc Update( [](http://example.com/codeflow?start=1&length=11)

why is this adding 3 different kinds of objects at the same time? Shouldn't they be managed independently?

cmd/root/install/mssql/server.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package mssql
```
shueybubbles, line 4: >e mssql [](http://example.com/codeflow?start=6&length=7)

this package name could be confusing if some .go file ever needs to reference both this one and the mssql package in github.com/microsoft/go-mssqldb

cmd/helpers/output/output.go
```diff
@@ -0,0 +1,202 @@
...
+	output.Info("Something noteworthy happened!")
+	output.Warn("You should probably take a look at this.")
+	output.Error("Something failed but I'm not quitting.")
+	// Calls os.Exit(1) after logging
+	output.Fatal("Bye.")
+	// Calls panic() after logging
+	output.Panic("I'm bailing.")
+*/
+
+package output
```
shueybubbles, line 16: >output [](http://example.com/codeflow?start=8&length=6)

do you expect the implementation packages might need to use this same output framework? If so it probably belongs in /internal or /pkg

This is a good time to establish rules for code layout to align with application logic and layering.

cmd is on top, pkg under that, internal under that. No layer can reference a layer above it.

shueybubbles: we could also have a contract with stuff under /pkg to accept io.Writer for output and errors and have the /cmd implementations pass the appropriate implementation down.

cmd/helpers/mssql/mssql.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+	"github.com/microsoft/go-sqlcmd/cmd/sqlconfig"
+	"github.com/microsoft/go-sqlcmd/pkg/sqlcmd"
+	"os"
+)
+
+func Connect(endpoint sqlconfig.Endpoint, user sqlconfig.User, console sqlcmd.Console) *sqlcmd.Sqlcmd {
+	// BUG(stuartpa): Check what the value of fromEnvironment should be here
```
shueybubbles, line 14: >(stuartpa): Check what the valu [](http://example.com/codeflow?start=7&length=31)

Using false or true depends on whether you want sqlcmd to pick up variables for formatting the output.

cmd/helpers/mssql/mssql.go
```diff
@@ -0,0 +1,38 @@
...
+	v := sqlcmd.InitializeVariables(false)
+	s := sqlcmd.New(console, "", v)
+	s.Format = sqlcmd.NewSQLCmdDefaultFormatter(false)
```
shueybubbles, line 17: >ewSQLCmdDefaultFormatter(false) [](http://example.com/codeflow?start=20&length=31)

opportunity here to provide a JSON formatter!

shueybubbles: (is there a good standard for encoding arbitrary data tables as json?)

cmd/helpers/mssql/mssql.go
```diff
@@ -0,0 +1,38 @@
...
+	connect := sqlcmd.ConnectSettings{
+		ServerName: fmt.Sprintf(
+			"%s,%d",
+			endpoint.EndpointDetails.Address,
+			endpoint.EndpointDetails.Port),
+		UseTrustedConnection: false,
```
shueybubbles, line 23: >seTrustedConnection: false, [](http://example.com/codeflow?start=3&length=27)

eventually user.UserDetails has to support integrated and AAD auth types, correct?

cmd/sqlconfig/sqlconfig.go
```diff
@@ -0,0 +1,49 @@
...
+	Endpoint string `mapstructure:"endpoint"`
+	User     string `mapstructure:"user"`
+}
+
+type Context struct {
+	ContextDetails `mapstructure:"context" yaml:"context"`
+	Name           string `mapstructure:"name"`
+}
+
+type UserDetails struct {
```
shueybubbles, line 32: >pe UserDetails s [](http://example.com/codeflow?start=2&length=16)

this seems "basic auth" specific
Perhaps User needs an authentication type value and a map that holds parameters specific to the auth type.

Where do we hook up the -U switch on the CLI as an override of the user name of the current context?

cmd/helpers/doc.go
stuartpa: At the moment, if it returns err I'm checking it with a checkErr(err) call, and letting the injected handler make a decision on whether to ignore it.  If we get a non nil err I think I want to know about it for now, let's find out what unexpected things fail.

cmd/helpers/mssql/mssql.go
stuartpa: We (SQL Server) built "FOR XML", and then "FOR JSON" clauses (for SELECT statement) that encode result sets into XML/JSON, I presume we should clone that functionality here

stuartpa:
> 

Done

---
In reply to: [1279168195](https://github.com/microsoft/go-sqlcmd/pull/138#issuecomment-1279168195) [](http://example.com/codeflow?ancestors=1279168195,1279168195)

---
Refers to: .gitignore:27 in 3db8294. [](commit_id = 3db829431465689d481d049b4b091272c20168f3, deletion_comment = False)

cmd/helpers/mssql/mssql.go
stuartpa: Ok, I've set to True

cmd/cmd.go
stuartpa: Done in next iteration

cmd/defaults_bcp.go
stuartpa: Done in next iteration

cmd/cmd.go
stuartpa: Done, although I am a "only one return" per function person.  But we should thrash this out in coding guidelines later.

cmd/helpers/mssql/mssql.go
stuartpa: Added issue:

[https://github.com/microsoft/go-sqlcmd/issues/151](https://github.com/microsoft/go-sqlcmd/issues/151)

cmd/helpers/mssql/mssql.go
stuartpa: Yes, completely agree.  I presume with docker installs no-one uses a trusted connection, hence the hardcode here.  But you raise a good point that this impl is focused on sql auth only at the moment, the 99.99% case I believe for docker.

build/azure-pipelines/build-common.yml
stuartpa: I have everything turned off on my IDE.  The only thing I did for this commit was run `gofmt -w .`, I specifcally had everything turned off in my IDE, and ran gofmt for this one commit to track down this issue.  I'll investigate further.

cmd/helpers/output/output.go
```diff
@@ -0,0 +1,202 @@
...
+		panic("formatter not initialized")
+	}
+
+	formatter.Serialize(in)
+}
+
+func Tracef(format string, a ...any) {
+	if loggingLevel >= verbosity.Trace {
+		format = ensureEol(format)
+		_, err := fmt.Printf("TRACE: ")
```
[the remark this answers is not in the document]
stuartpa: Good feedback, I've decoupled the output package from fmt.Printf, user of package has to pass in a io.WriteCloser (which will usually be os.Stdout / os.Stderr)

cmd/helpers/config/config.go
stuartpa: I've renamed the method to AddContextWithContainer.  These are all the input parameters needed to add a context which has an endpoint which has a docker container which is running SQL with basic auth on an endpoint.  This is a lot of params for a function, Scott Meyers would probably suggest adding an "Info" class here to carry the parameters.  Not sure if that's overkill for this single call.

cmd/helpers/config/config.go
stuartpa: Agree, good feedback, renamed to "AddContextWithContainer"

cmd/root/install/mssql/server.go
stuartpa: Each package in this folder structure is the name of the command line, so this is the .go file/package for `sqlcmd install mssql`.  This is the top most package in the architecture, so no-one will use this package.  But it does take a dependency on the /internal/helpers/mssql package, which is wrapping pkg/sqlcmd for error handling and tracing when we connect to sql to disable the sa account.   This is a lot of packages named mssql, let me think about it.  I think if there are no layering violations there will never be more than one needed in scope.

cmd/helpers/mssql/mssql.go
stuartpa: I've added authentication-type to sqlconfig, and this PR works for basic and trusted.  Added an issue for AAD auth, which I'll impl straight after this PR

go.mod
stuartpa: Made this an opt-in on windows, with the --encrypt-password switch

build/azure-pipelines/build-common.yml
stuartpa: The whitespace issue is main is using CRLF and the files I am editing git is converting to LF.

I presume we want to be a LF based repo.  Agree?

If so I can make a single PR that LFs the whole repo.

shueybubbles: I just have this in my git client config

```
core.autocrlf=true
```
Why would we not want to use that?

shueybubbles: Besides the test files for sqlcmd need to render with `\r\n` when testing on Windows and `\n` when testing on Linux.

stuartpa: The issue with requiring folks editing their git client is ensuring everyone does it (and this doesn't seem to be the default for many installs).

stuartpa: We've fixed this with a .gitattributes file

cmd/sqlcmd/main.go
```diff
@@ -1,278 +1,281 @@
...
+			return azuread.ActiveDirectoryInteractive
+		}
+	}
+	if a.AuthenticationMethod == "" {
+		return sqlcmd.NotSpecified
+	}
+	return a.AuthenticationMethod
+}
+
+//func main() {
```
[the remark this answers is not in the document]
stuartpa: I've added the cmd/modern structure with it's own main.go

cmd/modern/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	outputter = output.New(output.Options{StandardWriter: buf})
+	displayHints([]string{"This is a hint"})
+	assert.Equal(t, pal.LineBreak()+
+		"HINT:"+
+		pal.LineBreak()+
+		"  1. This is a hint"+pal.LineBreak(), buf.String())
+}
+
+func TestCheckErr(t *testing.T) {
+	defer func() { test.CatchExpectedError(recover(), t) }()
```
shueybubbles, line 40: >defer func() { test.CatchExpectedError(recover(), t) }() [](http://example.com/codeflow?start=1&length=56)

does the assert library not have a way to check for a method call leading to a panic?

cmd/modern/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/microsoft/go-sqlcmd/internal/pal"
+	"github.com/microsoft/go-sqlcmd/internal/test"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+func TestMainStart(t *testing.T) {
+	os.Args[1] = "--help"
+	main()
```
shueybubbles, line 20: >main() [](http://example.com/codeflow?start=1&length=6)

"doesn't panic" seems like a very minimal test. Couldn't this test verify that rootCmd is constructe with proper values?

cmd/modern/root_test.go
```diff
@@ -0,0 +1,27 @@
...
+	c.Execute()
+}
+
+func TestIsValidSubCommand(t *testing.T) {
+	c := cmdparser.New[*Root](dependency.Options{})
+	invalid := c.IsValidSubCommand("nope")
+	assert.Equal(t, false, invalid)
+	valid := c.IsValidSubCommand("query")
+	assert.Equal(t, true, valid)
+}
```
shueybubbles, line 27: using Equalf with a message would improve logging. If a test fails it'd be nice to know which assertion was failing without having to get the line number and looking up the code. Line numbers are ephemeral too so if you are looking at a test failure you have to know which branch/commit it was synced to.

internal/cmdparser/test.go
```diff
@@ -0,0 +1,60 @@
+package cmdparser
+
+import (
```
shueybubbles, line 3: the existence of both "test.go" and "test_test.go" is breaking my brain.

internal/secret/secret_test.go
```diff
@@ -4,67 +4,25 @@
...
-			wantPlainText: "",
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
+func TestEncodeAndDecode(t *testing.T) {
+	notEncrypted := Encode("plainText", false)
+	encrypted := Encode("plainText", true)
+	Decode(notEncrypted, false)
+	Decode(encrypted, true)
```
shueybubbles, line 15: shouldn't this assert the decoded string is correct?

internal/pal/intialize.go
```diff
@@ -8,9 +8,10 @@ func init() {
 		if err != nil {
 			panic(err)
 		}
-	})
+	}, "\n")
```
shueybubbles, line 11: >\n") [](http://example.com/codeflow?start=5&length=4)

might Windows users want \r\n ?

shueybubbles:
the structure looks ok to me. I don't see much value in unit tests that exercise a function with no assertions about the behavior. A unit test should serve as a specification (code > English) for a method, but without assertions there's no definition of correct behavior other than "it doesn't panic".  With the dependency structure you have established, unit tests should have an easy ability to inject an interface implementation into the dependencies and verify that the method under test utilized that dependency correctly, much the way C# tests do with Moq.  That way the unit test establishes the relationship between a given method and its dependencies and can truly prevent regressions in the future.

shueybubbles (approved):
:shipit:

cmd/modern/main_test.go
stuartpa: Yes there is, the assert.Panics* functions

https://pkg.go.dev/github.com/stretchr/testify/assert#Panics

I've started a follow-up PR, will replace this pattern with assert.Panics

internal/cmdparser/test.go
stuartpa: I've added comments to both of these files.  test.go provides functions to write compact tests of the CLI application using this cmdparser package, and test_test.go tests these test function (testing the tests).  test_test.go is useful also, because it shows how easy it is to create a CLI using the cmdparser package.

internal/pal/intialize.go
stuartpa: Added a defaultLineBreak function that returns \n or \r\n depending on platform

internal/secret/secret_test.go
stuartpa: Agree, I have leaned to much on adding unit tests for the code coverage only, without doing enough validation.  I've started a new PR to ensure all the unit tests include validation, so we will have good code coverage with validating results as well.

I see the testify package also supports mocking, I will look at using the mocking functionality to speed up the unit tests around the install/uninstall functionality.

PR #142, opened 2022-10-03
block sys commands on -X

shueybubbles:
I can't quite figure out how to make `-X [1]` work with Kong, so for now I'm staying with `-X`  and warning when a sys command used.


stuartpa (approved):
:shipit:

PR #143, opened 2022-10-06
Redirect sqlcmd errors to err stream

apoorvdeshmukh:
By default sqlcmd uses stdout as the default stream for output and errors.
As a result when the stderr is redirected externally to file, the sqlcmd errors are missed.
This commit logs all the sqlcmd errors to the stderr of OS in case default output
stream is stdout and stderr for sqlcmd is not set.

### DIFF
--- a/pkg/sqlcmd/sqlcmd.go
+++ b/pkg/sqlcmd/sqlcmd.go
@@ -136,7 +136,7 @@ func (s *Sqlcmd) Run(once bool, processAll bool) error {
 				args = make([]string, 0)
 				once = true
 			} else {
-				_, _ = s.GetOutput().Write([]byte(err.Error() + SqlcmdEol))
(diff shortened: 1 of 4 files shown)


pkg/sqlcmd/sqlcmd.go
```diff
@@ -209,6 +209,19 @@ func (s *Sqlcmd) SetError(e io.WriteCloser) {
 	s.err = e
 }
 
+// WriteError writes the error on specified stream
+func (s *Sqlcmd) WriteError(stream io.Writer, err error) {
```
shueybubbles, line 213: I like the consolidation into this function

shueybubbles (changes requested):
:clock1:

shueybubbles (approved):
:shipit:

Issue #140, opened 2022-09-21
Provide useful HINT to user who passes -P into command line


stuartpa:
C:\Users\user>sqlcmd -U sa -P blahblah
sqlcmd.exe: error: unknown flag -P, did you mean one of "-c", "-C", "-d", "-E", "-U", "-i", "-o", "-q", "-Q", "-S", "-X", "-G", "-x", "-v", "-a", "-l", "-H", "-K", "-N", "-b", "-V", "-m", "-F", "-r", "-h", "-u", "-?"?

Let the user know in this "sqlcmd: error:" that this version of sqlcmd requires -P to be passed in as an environment variable, which I can never remember off the top of my head, because MSSQL on Linux used SA_PASSWORD, which is now MSSQL_SA_PASSWORD, but for SQLCMD it's SQLCMDPASSWORD.

PR #152, opened 2022-10-27
Stub -M and fix !!

shueybubbles:
Customer feedback from a financial services company - 
`!!` shouldn't require a space after it, and `-M` should just be ignored.


cmd/sqlcmd/main.go
```diff
@@ -58,6 +58,7 @@ type SQLCmdArguments struct {
 	ColumnSeparator             string            `short:"s" help:"Specifies the column separator character. Sets the SQLCMDCOLSEP variable."`
 	ScreenWidth                 *int              `short:"w" help:"Specifies the screen width for output. Sets the SQLCMDCOLWIDTH variable."`
 	TrimSpaces                  bool              `short:"W" help:"Remove trailing spaces from a column."`
+	MultiSubnetFailover         bool              `short:"M" help:"Provided for backward compatibility."`
```
stuartpa, line 61: Is it possible to add a long: parameter to give a bit more detail on what it means to be "provided for backward compat".... in the long we say this option is a no-op, does nothing, and it provided to not break existing scripts.  The short: right now doesn't tell the user that this does nothing.

stuartpa (approved):
:shipit:

shueybubbles:
Fixes #140

Issue #154, opened 2022-10-29
Add a command line canonicalizer to internally mask funky back-compat switches  


stuartpa:
Some of the original sqlcmd (ODBC base) command line switches are completely non-standard by modern standards, such as  -x1 / -k1 -k2 etc and therefore not supportable by any modern CLI framework (such as Kong or Cobra), rather than try to hack these CLI frameworks, add a canonicalizer very early in process start-up that looks for these odd command line switches and re-writes them into a standard form, and then passes args in standard form into the CLI framework Cobra/Kong.

stuartpa:
Another PR comment related to this, recording here:

"i suspect some version of this code is permanent. We aren't likely to reproduce all the legacy switches in the new CLI are we? Perhaps we should use an environment variable to only allow modern mode where it uses query as the default command but only with whatever switches that command supports." - David

"I think we can solve the problem of the legacy switches by writing a small "canonicalizer" called at the start of main, which scans through the supplied args, and modifies them to be parsable by Cobra, then call cobra.command.SetArgs to override what's passed in from env." - Stuart

shueybubbles:
i think we can make these work with Cobra without a special case.

PR #165, opened 2022-11-22
Modern POSIX style CLI for SQLCMD

stuartpa:
Main package (main.go) is the entry point for the sqlcmd CLI application.

To follow the flow of this code:

 1. enter through main.go, (TEMPORARY: decision made whether to invoke the modern
    cobra CLI
 2. Then cmd/cmd.go, see the init() func `New` the `Root` cmd (and all its
    subcommands)
 3. The command-line is then parsed and internal.Initialize() runs (with
    the logging level, config file path, error handling and trace support passed
    into internal packages)
 4. Now go to the cmd/root/… folder structure, and read the DefineCommand
    function for the command (sqlcmd install, sqlcmd query etc.) being run
 5. Each cmd/root/... command has a `run` method that performs the action
 6. All the commands (cmd/root/…) use the /internal packages to abstract from error
    handling and trace (non-localized) logging (as can be seen from the `import`
    for each command (in /cmd/root/...)).

This code follows the Go Style Guide

  - https://google.github.io/styleguide/go/
  - https://go.dev/doc/effective_go
  - https://github.com/golang-standards/project-layout


main.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/microsoft/go-sqlcmd/cmd"
+	legacyCmd "github.com/microsoft/go-sqlcmd/cmd/sqlcmd"
+	"os"
+)
+
+// main is the entrypoint function for sqlcmd.
+//
+// TEMPORARY: While we have both the new cobra and old kong CLI
+// implementations, main decides which CLI framework to use
```
shueybubbles, line 16: i suspect some version of this code is permanent. We aren't likely to reproduce all the legacy switches in the new CLI are we? Perhaps we should use an environment variable to only allow modern mode where it uses `query` as the default command but only with whatever switches that command supports.

cmd/cmd.go
```diff
@@ -0,0 +1,79 @@
...
+
+func initialize() {
+	options := internal.InitializeOptions{
+		ErrorHandler: checkErr,
+		HintHandler:  displayHints,
+		OutputType:   "yaml",
+		LoggingLevel: 2,
+	}
+
+	config.SetFileName(configFilename)
```
shueybubbles, line 36: >configFilename [](http://example.com/codeflow?start=20&length=14)

configFilename isn't assigned any value. Is it supposed to be nul for the default with an option for the user to pass it along as a command line parameter or environment variable?

cmd/cmd.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+	rootCmd.CheckErr(err)
+}
+
+// displayHints displays helpful information on what the user should do next
+// to make progress.  displayHints is injected into dependencies (helpers etc.)
+func displayHints(hints []string) {
+	if len(hints) > 0 {
+		output.Infof("\nHINT:")
```
shueybubbles, line 74: >\n [](http://example.com/codeflow?start=16&length=2)

are we going to use \n on all platforms? Do all the output renderers on Windows do the right thing with it?

cmd/cmd_test.go
```diff
@@ -0,0 +1,315 @@
...
+		t.Run(tt.name, func(t *testing.T) { runTests(t, tt) })
+	}
+
+	verifyConfigIsEmpty(t)
+}
+
+func verifyConfigIsEmpty(t *testing.T) {
+	if !config.IsEmpty() {
+		bytes := output.Struct(config.GetRedactedConfig(true))
+		t.Errorf("Config is not empty. Content of config file:\n%s\nConfig file used:%s",
```
shueybubbles, line 276: >t.Errorf [](http://example.com/codeflow?start=2&length=8)

github.com/stretchr/testify/assert is a pretty good assert library

cmd/cmd_test.go
```diff
@@ -0,0 +1,315 @@
...
+	// If test name starts with 'neg-' expect a Panic
+	if strings.HasPrefix(tt.name, "neg-") {
+		defer func() {
+			if r := recover(); r == nil {
+				t.Errorf("The code did not panic")
+			}
+		}()
+		cmd.Execute()
+	}
+	cmd.Execute()
```
shueybubbles, line 234: is silence considered success? The lack of asserts makes it hard to tell what's being tested.

internal/cmdparser/cmd.go
```diff
@@ -0,0 +1,208 @@
...
+
+// CheckErr passes the error down to cobra.CheckErr (which is likely to call
+// os.Exit(1) if err != nil.  Although if running in the golang unit test framework
+// we do not want to have os.Exit() called, as this exits the unit test runner
+// process, and call panic instead so the call stack can be added to the unit test
+// output.
+func (c *Cmd) CheckErr(err error) {
+	// If we are in a unit test driver, then panic, otherwise pass down to cobra.CheckErr
+	if strings.HasSuffix(os.Args[0], ".test") || // are we in go test?
+		(len(os.Args) > 1 && os.Args[1] == "-test.v") { // are we in goland unittest?
```
shueybubbles, line 122: this seems like a flimsy convention considering someone else could write a new test framework etc. Can these test frameworks not set some global state instead of relying on command line parsing?

cmd/sqlconfig/doc.go
```diff
@@ -0,0 +1,58 @@
...
+# Example
+
+An example of the sqlconfig file looks like this:
+
+	apiversion: v1
+	endpoints:
+	- container:
+		id: 0e698e65e19d9c
+		image: mcr.microsoft.com/mssql/server:2022-latest
+	  endpoint:
```
shueybubbles, line 31: > endpoint: [](http://example.com/codeflow?start=2&length=10)

Let's pretend we will have scenarios involving non-TDS endpoints (eg PowerBI, postgres, SQL AS/RS/IS). Will adding those just be a matter of adding an optional "type" field to "endpoint" to identify it as needing a different driver? 

theoretically `sqlcmd query "some query"` could run any arbitrary query language if the query command were to route it to a package that knows how to connect to the non-TDS endpoint.

cmd/sqlconfig/sqlconfig.go
```diff
@@ -0,0 +1,55 @@
...
+	Address string `mapstructure:"address"`
+	Port    int    `mapstructure:"port"`
+}
+
+type ContainerDetails struct {
+	Id    string `mapstructure:"id"`
+	Image string `mapstructure:"image"`
+}
+
+type AssetDetails struct {
```
shueybubbles, line 16: >t { [](http://example.com/codeflow?start=23&length=3)

where does one find instructions for using `mapstructure` and `yaml` attributes?

internal/doc.go
```diff
@@ -0,0 +1,54 @@
...
+     e.g. Do not sprinkle application (non-helper) code with:
+       err, _ := fmt.printf("Hope this works")
+       if (err != nil) {
+         panic("How unlikely")
+       }
+     Do use the injected checkErr callback and let the application decide what to do
+       err, _ := printf("Hope this works)
+       checkErr(err)
+ - Do not have an internal package take a dependency on another internal package
+   unless they are building on each other, instead inject the needed capability in the
```
shueybubbles, line 33: >hey are building o [](http://example.com/codeflow?start=11&length=18)

this is very Electron-ish. Given the lack of a MEF equivalent it seems good.

internal/output/output.go
```diff
@@ -0,0 +1,208 @@
...
+	"io"
+	"regexp"
+	"strings"
+)
+
+var formatter Formatter
+var loggingLevel verbosity.Enum
+var runningUnitTests bool
+
+var standardWriteCloser io.WriteCloser
```
shueybubbles, line 35: >andardWriteCloser io.WriteCloser [](http://example.com/codeflow?start=6&length=32)

don't we need separate error and stdout channels? Most apps don't write everything to stdout, they use stderr too.

Legacy Sqlcmd also lets the user change where stdout and stderr write during app execution. Do you expect we'll keep that functionality with the new API such that this channel needs to support dynamic reassignment?

internal/output/formatter/xml.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+type Xml struct {
+	Base
+}
+
+func (f *Xml) Serialize(in interface{}) (bytes []byte) {
+	var err error
+
+	bytes, err = xml.MarshalIndent(in, "", "    ")
```
shueybubbles, line 17: >    ") [](http://example.com/codeflow?start=41&length=6)

2 spaces of indent should be sufficient

internal/output/verbosity/enum.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package verbosity
+
+type Enum int
```
shueybubbles, line 6: >Enum [](http://example.com/codeflow?start=5&length=4)

how about `Level` or even `Verbosity` ? `Enum` is the name of the mechanism, not the semantic of the type.

internal/pal/pal.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// CmdLineWithEnvVars generates a command-line that can be run at the
+// operating system command-line (e.g. bash or cmd) which also requires
+// one or more environment variables to also be set
+func CmdLineWithEnvVars(vars []string, cmd string) string {
+	var sb strings.Builder
+	for _, v := range vars {
+		sb.WriteString(envVarCommand())
+		sb.WriteString(cliQuoteIdentifier() + v + cliQuoteIdentifier())
```
shueybubbles, line 36: is there really value in splitting these up instead of having a single function that takes the variable name? 
This function could accept a map instead of an array, and pass the keys as the var names and the values as the var values. The pal function that exports the environment variable would parse the values and do any escaping of special characters.

internal/secret/generate.go
```diff
@@ -0,0 +1,60 @@
...
+	remainingLength := passwordLength - minSpecialChar - minNum - minUpperCase
+	for i := 0; i < remainingLength; i++ {
+		idx, err := rand.Int(rand.Reader, big.NewInt(int64(len(allCharSet))))
+		checkErr(err)
+		_, err = password.WriteString(string(allCharSet[idx.Int64()]))
+		checkErr(err)
+	}
+
+	inRune := []rune(password.String())
+	mathRand.Shuffle(len(inRune), func(i, j int) {
```
shueybubbles, line 56: interview question - implement this function!

internal/pal/intialize.go
```diff
@@ -0,0 +1,16 @@
...
+
+func init() {
+	Initialize(func(err error) {
+		if err != nil {
+			panic(err)
+		}
+	})
+}
+
+func Initialize(handler func(err error)) {
```
shueybubbles, line 14: >handler func(err error) [](http://example.com/codeflow?start=16&length=23)

would it be a bit cleaner to define an interface or type for this handler func instead of using raw `func(err error)` everywhere?

internal/net/net.go
```diff
@@ -0,0 +1,42 @@
...
+	timeout := time.Second
+	trace(
+		"Checking if local port %d is available using DialTimeout(tcp, %v, timeout: %d)",
+		port,
+		net.JoinHostPort("localhost", strconv.Itoa(port)),
+		timeout,
+	)
+	conn, err := net.DialTimeout(
+		"tcp",
+		net.JoinHostPort("localhost", strconv.Itoa(port)),
```
shueybubbles, line 22: >net.JoinHostPort("localhost", strconv.Itoa(port)), [](http://example.com/codeflow?start=2&length=50)

just call this once and pass it to both trace and dial

cmd/root/config/add-user.go
```diff
@@ -0,0 +1,107 @@
...
+			"The --encrypt-password flag can only be used when authentication type is 'basic'")
+	}
+
+	user := sqlconfig.User{
+		Name:               c.name,
+		AuthenticationType: c.authType,
+	}
+
+	if c.authType == "basic" {
+		if os.Getenv("SQLCMD_PASSWORD") == "" {
```
shueybubbles, line 84: >	if os.Getenv("SQLCMD_PASSWORD") == "" { [](http://example.com/codeflow?start=1&length=40)

we should prompt for the password on the console.
we could also allow reading it from redirected stdin

`sqlcmd config add-user --username myuser < somefile.txt`

cmd/root/config/connection-strings.go
```diff
@@ -0,0 +1,68 @@
...
+		Aliases: []string{"cs"},
+	}
+
+	c.Cmd.DefineCommand()
+}
+
+func (c *ConnectionStrings) run() {
+	// connectionStringFormats borrowed from "portal.azure.com" "connection strings" pane
+	var connectionStringFormats = map[string]string{
+		"ADO.NET": "Server=tcp:%s,%d;Initial Catalog=%s;Persist Security Options=False;User ID=%s;Password=%s;MultipleActiveResultSets=False;Encode=True;TrustServerCertificate=False;Connection Timeout=30;",
```
shueybubbles, line 41: >Encode=True [](http://example.com/codeflow?start=135&length=11)

what is "Encode" ?

cmd/root/install/mssql-base.go
```diff
@@ -0,0 +1,324 @@
...
+		DefaultString: "SQL_Latin1_General_CP1_CI_AS",
+		Name:          "collation",
+		Usage:         "The SQL Server collation",
+	})
+}
+
+func (c *MssqlBase) Run() {
+	var imageName string
+
+	if !c.acceptEula && viper.GetString("ACCEPT_EULA") == "" {
```
shueybubbles, line 159: >if !c.acceptEula && viper.GetString("ACCEPT_EULA") == " [](http://example.com/codeflow?start=1&length=55)

viper has a way to bind its config values to flags. Do you see us using that technique such that code like this would only have to check in viper?

shueybubbles (approved):
:shipit:

cmd/cmd.go
stuartpa: It's set the default value for the --sqlconfig flag, so it's ~\.sqlcmd\sqlconfig (this func is called by the cobra framework after the flag defaults have been set

cmd/root/config/add-user.go
stuartpa: Added issue and assigned to myself (to break this PR up a bit):

[https://github.com/microsoft/go-sqlcmd/issues/167](https://github.com/microsoft/go-sqlcmd/issues/167)

main.go
stuartpa: I think we can solve the problem of the legacy switches by writing a small "canonicalizer" called at the start of main, which scans through the supplied args, and modifies them to be parsable by Cobra, then call cobra.command.SetArgs to override what's passed in from env.

I'll look at this staight after this PR is in.  I've entered an issue and assigned it to myself.

cmd/cmd.go
stuartpa: Good question.  I don't know if all the output renderers on Windows do the right thing.  Should i replace this with `SqlcmdEol` to be safe?

main.go
stuartpa: Added to:

[https://github.com/microsoft/go-sqlcmd/issues/154](https://github.com/microsoft/go-sqlcmd/issues/154)

cmd/cmd_test.go
stuartpa: Agree, I was focusing on getting to 100% code coverage, but now I need to go back and actually validate the test results.  To keep the size of the PR managable I'll follow up now with a Test focused PR.

internal/cmdparser/cmd.go
stuartpa: Added issue to track

[https://github.com/microsoft/go-sqlcmd/issues/168](https://github.com/microsoft/go-sqlcmd/issues/168)

internal/output/verbosity/enum.go
stuartpa: I get what you're saying, but I've done it this way so it looks like this when using it:

 output.New(verbosity.Enum(options.LoggingLevel))

golang uses the package name to add the context. 

But agree this might be me trying to make it familiar to a C# user.

I'll change it to verbosity.Level

internal/output/formatter/xml.go
stuartpa: Done for both this one (XML) and Json

internal/secret/generate.go
stuartpa: :smile:

internal/output/output.go
stuartpa: I started with two channels, but then didn't know what semantics to apply to do the separation, so I removed stderr to keep this PR contained.  I've added an Issue, and assigned to myself:

[https://github.com/microsoft/go-sqlcmd/issues/169](https://github.com/microsoft/go-sqlcmd/issues/169)

internal/net/net.go
stuartpa: Done (added to the follow on PR in shueybubbles/bcp)

internal/doc.go
stuartpa: :thumbsup:

cmd/root/install/mssql-base.go
stuartpa: Is I am using the viper feature to bind to env here ( err = viper.BindEnv("ACCEPT_EULA")), but no binding to the deep sub cmd flag.  The reason I didn't do that is the flag is deep down in the hierarchy, e.g. sqlcmd install mssql --flag, which would be a long variable name, someting like SQLCMD_INSTALL_MSSQL_ACCEPT_EULA I think)

cmd/root/config/connection-strings.go
stuartpa: Added more general issue to review these connection strings:

"Review sqlcmd config connection-strings"

[https://github.com/microsoft/go-sqlcmd/issues/170](https://github.com/microsoft/go-sqlcmd/issues/170)

internal/pal/pal.go
stuartpa: The value is code coverage reporting :smile: .  I'd like to promote a culture that a lot of the files in this repo are at 100% code coverage ("most" already files are),  (I want to add a "100PercentClub" build pipeline check so a list (or source file metadata tag?) of all files that are at 100% code coverage to never drop below 100%).

There is 2nd reason, the other approach is to use runtime.GOOS == "windows" etc., but this generates a style failure in go vet (and IDEs), because it is flagged as "unreachable" so you never get the little "green tick" for the source file being 100% good.

internal/pal/intialize.go
stuartpa: I've added a more general issue to cover this area of style guidelines.   I think I agree with you (this is what I'd do in c#), but would using interface encourage tighter coupling.  Is the idomatic golang apporach for the caller to define the interface.

stuartpa: [https://github.com/microsoft/go-sqlcmd/issues/171](https://github.com/microsoft/go-sqlcmd/issues/171)

internal/pal/pal.go
stuartpa: I've added issue:

[https://github.com/microsoft/go-sqlcmd/issues/172](https://github.com/microsoft/go-sqlcmd/issues/172)

Protect source files already at 100% code coverage - add "100PercentClub" build pipeline check to

cmd/sqlconfig/doc.go
stuartpa: I love the growth mind set here! :smile:

internal/pal/pal.go
stuartpa: I've added comment to:

[https://github.com/microsoft/go-sqlcmd/issues/171](https://github.com/microsoft/go-sqlcmd/issues/171)

"Another topic around coding guidelines around Platform Abstraction, should we use runtime.GOOS == "windows" etc. or the filename_windows, filename_darwin, filename_linux approach:"

cmd/sqlconfig/doc.go
stuartpa: Related to this, to be able the best connection strings from:

sqlcmd config connection-strings

Do we need to keep some metadata for the endpoint that tells us it is mssql-server, or mssql-edge, or azsqldb, or postgres! So we know how to generate a good connection string for it

stuartpa: Added:

[https://github.com/microsoft/go-sqlcmd/issues/173](https://github.com/microsoft/go-sqlcmd/issues/173)

Should ~/.sqlcmd/sqlconfig keep metadata on what the endpoint is

cmd/cmd_test.go
stuartpa: Added to:

[https://github.com/microsoft/go-sqlcmd/issues/171](https://github.com/microsoft/go-sqlcmd/issues/171)


Should tests standarize on a specific assert library e.g. github.com/stretchr/testify/assert

Also in the Go Style Guide, what does "invalid local style" / "Use of assertion-based testing libraries" mean:

[https://google.github.io/styleguide/go/guide#local-consistency](https://google.github.io/styleguide/go/guide#local-consistency)

"
Examples of invalid local style considerations:

Line length restrictions for code
Use of assertion-based testing libraries
"

cmd/cmd.go
stuartpa: I've replaced all to sqlcmd.SqlcmdEol for next PR (in sheuybubble/bcp) (which I'll then need to work out how to inject, rather than having everything take an import on sqlcmd package)

internal/pal/pal.go
stuartpa: I'll refactor func to take a map in next PR

cmd/sqlconfig/sqlconfig.go
stuartpa: They're called "struct tags", and you're right to ask, because they are barely documented in the golang spec:

This answers to this stackoverflow question seems to do the best at gathering the links to learn more:
 
[https://stackoverflow.com/questions/10858787/what-are-the-uses-for-struct-tags-in-go](https://stackoverflow.com/questions/10858787/what-are-the-uses-for-struct-tags-in-go)

stuartpa:
Thanks for the all the PR feedback.  I'll follow up with a quick PR to neaten the unit tests. (I was focused on getting most files to 100% code coverage, need to go back and validate/assert result output). 

---
In reply to: [1196841042](https://github.com/microsoft/go-sqlcmd/pull/165#issuecomment-1196841042) [](http://example.com/codeflow?ancestors=1196841042,1196841042)

shueybubbles:
Could SqlcmdEol declaration move to internal?


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

internal/pal/pal.go
shueybubbles: I like using the `file_linux` (or build tags) approach so per-platform code isn't intermixed. 
I wasn't asking to merge all the functions into one using the `GOOS=="windows"` check, I just meant to merge the `envVarCommand()` and `cliQuoteIdentifier()` concatenations into a single function. That function would take both the name and the value of the variable so it can properly escape the value for the current platform.

PR #177, opened 2022-12-22
I18n and L10n support for go-sqlcmd

apoorvdeshmukh:
**Brief:**
This is a draft PR to demonstrate how i18n and l10n support can be added for go-sqlcmd.
$go-sqlcmd\internal\translations\translations.go contains the command for extracting strings to be translated.
Strings that the application would localize can be marked by using translate() function from localizer package.

```
PS git\go-sqlcmd> $env:SQLCMD_LOCALE = 'fr-ch'
PS git\go-sqlcmd> .\sqlcmd.exe -a 2
sqlcmd.exe: error: sqlcmd.exe: '-a 2': La taille du paquet doit être un nombre compris entre 512 et 32767.
PS git\go-sqlcmd> $env:SQLCMD_LOCALE = ''     
PS git\go-sqlcmd> .\sqlcmd.exe -a 2      
sqlcmd.exe: error: sqlcmd.exe: '-a 2': Packet size has to be a number between 512 and 32767.
```

**General Steps for generating and updating translations.**
1. Run 'go generate .\internal\translations\translations.go' // This generates/updates out.gotext.json files under under .\internal\translations\locales\<locale>\ directory which can be provided for translator. It even warns in case there are missing translations.
2. The out.gotext.json files need to be submitted to translator and once translated needs to be copied in same directory as messages.gotext.json
3. Run 'go generate .\internal\translations\translations.go' to ensure there are no warnings for missing translations.
4. set locale by setting env variable SQLCMD_LOCALE to a BCP47 language tag in lowercase.

go-sqlcmd should now display the messages in localized manner.

**Next Steps:**
I propose that we consolidate all the strings in go-sqlcmd to a single file to keep track of them and to also determine what should be localized and what should not be.
Once this is done, out.gotext.json files will be generated and updated.

**Pending Action Items:**
We need to discuss and finalize how the process of translation would be carried out and what would be the guidelines for community.


apoorvdeshmukh:
The test failures are due to the missing internal pkg localizer which is also being added as part of this PR.
Current PR will only be used to showcase E2E changes.
I will split this PR. 
1st PR to only add the localizer package. This will ensure that localizer package is available for consumption.
2nd PR to showcase consumption of localizer package to demonstrate translation workflow as mentioned in this PR.

cmd/sqlcmd/sqlcmd.go
```diff
@@ -10,11 +10,14 @@ import (
 
 	"github.com/alecthomas/kong"
 	"github.com/microsoft/go-mssqldb/azuread"
+	"github.com/microsoft/go-sqlcmd/internal/localizer"
```
shueybubbles, line 13: >-sqlcmd/internal/localizer" [](http://example.com/codeflow?start=25&length=27)

there's no localizer in the PR

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,7 +121,15 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
 	return a.AuthenticationMethod
 }
 
+func initLocale() {
+	localeName := os.Getenv("SQLCMD_LOCALE")
```
shueybubbles, line 125: >os.Getenv("SQLCMD_LOCALE") [](http://example.com/codeflow?start=15&length=26)

it's not a locale we are getting, it's a language.
Locale is about formatting based on regional settings.

how about SQLCMD_LANGID or SQLCMD_LANGUAGE
We could potentially support LCIDs too

shueybubbles: are there os-specific methods to get the user's current preferred language? Windows users can use different UI languages than the OS language.
The ODBC sqlcmd only installs 1 language, IIUC. We'll be installing all languages so we should be able to use a fallback logic that's aware of the Windows MUI model.

shueybubbles:
@stuartpa do you know if the loc team would be ok with these json files to translate? Maybe we should meet with Khoi Pham.
If they are amenable, this model looks fine to me.

internal/translations/locales/de-DE/out.gotext.json
```diff
@@ -0,0 +1,20 @@
+{
```
stuartpa, line 1: This contents of this file seems to be duplicate of messages.gotext.json, is this intentional?

internal/translations/catalog.go
```diff
@@ -0,0 +1,68 @@
+// Code generated by running "go generate" in golang.org/x/text. DO NOT EDIT.
```
stuartpa, line 1: Do we actually check this code in, or is it generated by the build and kept on the local machine only?

cmd/sqlcmd/sqlcmd.go
stuartpa: I see Localizer in internal/localizer folder

cmd/sqlcmd/sqlcmd.go
apoorvdeshmukh: Currently we expect BCP47 tag in lowercase but can be extended to LCIDs too. 
I did not consider windows MUI model. Will look into the window's specific APIs for the same if any.

internal/translations/locales/de-DE/messages.gotext.json
```diff
@@ -0,0 +1,20 @@
+{
+    "language": "de-DE",
+    "messages": [
+        {
+            "id": "'-a {PacketSize}': Packet size has to be a number between 512 and 32767.",
```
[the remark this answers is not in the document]
apoorvdeshmukh: String in src code will be wrapped with translate() function and the go generate command will update the id and generate catalog.go file whenever executed. So I think, even if strings are changed later, one only needs to run go generate command and keep the translation blank (if already filled) to be translated again by the central localization team.

internal/translations/catalog.go
apoorvdeshmukh: This code will automatically be generated by go generate command and is responsible for mapping strings to their translations. As we are checking in the translations, I thought of adding the generated catalog.go file as well.

apoorvdeshmukh:
In the last commit, I have marked strings which need to be localized. Need feedback on if there is anything else that needs to be localized. (I have omitted help text for modern CLI and kong help text as we are in process of replacing kong with cobra.)

PR #178, opened 2022-12-27
Improve unit test verification

stuartpa:
This PR is being merged as part of https://github.com/microsoft/go-sqlcmd/pull/205


cmd/modern/root/query.go
```diff
@@ -63,11 +63,12 @@ func (c *Query) run() {
 		line = console.NewConsole("")
 		defer line.Close()
 	}
-	s := mssql.Connect(endpoint, user, line)
+	sql := mssql.New(false)
```
shueybubbles, line 66: >ew(false) [](http://example.com/codeflow?start=15&length=9)

Boolean parameters are where I miss named parameters in C#. I have to look up the New function to figure out what is being set to false.

internal/mssql/factory.go
```diff
@@ -0,0 +1,9 @@
+package mssql
+
+func New(unitTesting bool) MssqlInterface {
```
shueybubbles, line 3: >fun [](http://example.com/codeflow?start=0&length=3)

it's a little weird to me for the production code and the test code to live in the same binary.

Could the mock implementation only live in a _test.go file and the factory accept an optional delegate used to create the instance? If no delegate is passed, it uses the default production type. 
The unit tests would pass a delegate that creates the mock.
Alternatively we could use a build tag to control inclusion of the mock type and make sure the official product build sets the tag to exclude it.

cmd/modern/root/query.go
stuartpa: Good point, in my IDE (Goland), I see:

	sql := mssql.New( unitTesting: false)
	
The IDE puts the variable name in the display.  But for code reviews and other IDEs that don't do this, it is not clear at all.

I'll move this to the Options pattern which is idiomatic for golang

stuartpa: Moved to an Options parameter in:

[https://github.com/microsoft/go-sqlcmd/pull/205](https://github.com/microsoft/go-sqlcmd/pull/205)

internal/mssql/factory.go
stuartpa: I've made a note to talk about this at our next sync.  I'm doing this to speed up unit testing (of other units), I think the mock impl needs to be available in prod code here, but let's chat

PR #180, opened 2022-12-29
Adding Global Flag in Modern CLI

JyotikaGargg:
In order to migrate all the commands from Kong to Modern CLI,
I have started with one flag here : InputFile.

### DIFF
--- a/cmd/modern/root.go
+++ b/cmd/modern/root.go
@@ -17,6 +17,7 @@ type Root struct {
 	configFilename string
 	loggingLevel   int
 	outputType     string
+	inputType      string
 }
 
 // DefineCommand defines the top-level sqlcmd sub-commands.
@@ -112,5 +113,13 @@ func (c *Root) addGlobalFlags() {
 		Name:       "verbosity",
 		Shorthand:  "v",
 		Usage:      "Log level, error=0, warn=1, info=2, debug=3, trace=4",
+	})
+
+	c.AddFlag(cmdparser.FlagOptions{
+		String:        &c.inputType,
+		DefaultString: "yaml",
+		Name:          "InputFile",
+		Shorthand:     "i",
+		Usage:         "output type (yaml, json or xml)",
 	})
 }


cmd/modern/root.go
```diff
@@ -17,6 +17,7 @@ type Root struct {
 	configFilename string
 	loggingLevel   int
 	outputType     string
+	inputType      string
```
shueybubbles, line 20: >inputType      string [](http://example.com/codeflow?start=1&length=21)

why is this a root level parameter?

cmd/modern/root.go
```diff
@@ -113,4 +114,12 @@ func (c *Root) addGlobalFlags() {
 		Shorthand:  "v",
 		Usage:      "Log level, error=0, warn=1, info=2, debug=3, trace=4",
 	})
+
+	c.AddFlag(cmdparser.FlagOptions{
+		String:        &c.inputType,
+		DefaultString: "yaml",
+		Name:          "InputFile",
+		Shorthand:     "i",
+		Usage:         "output type (yaml, json or xml)",
```
shueybubbles, line 123: this implies 3 different concepts - 
1. input type
2. input file name
3. Output type

shueybubbles:
The sqlcmd package defines the main parameter structure for the query command, here [https://github.com/microsoft/go-sqlcmd/blob/main/pkg/sqlcmd/connect.go](https://github.com/microsoft/go-sqlcmd/blob/main/pkg/sqlcmd/connect.go)

What I think you would do is update query.go here [https://github.com/microsoft/go-sqlcmd/blob/main/cmd/modern/root/query.go](https://github.com/microsoft/go-sqlcmd/blob/main/cmd/modern/root/query.go) to define the new modern command structure. That command implementation will construct a Connect object to pass to the Sqlcmd instance to do its work. 
`modern/root/query.go` will eventually replace `cmd/sqlcmd/sqlcmd.go`

cmd/modern/root.go
stuartpa: I think this is a copy and paste error (from outputType above)

cmd/modern/root.go
stuartpa: This is a good question.  Where should we put all these back-compat switches.  They are needed either when the customer uses the `query` subcommand, or when the use uses no subcommand (which is defined in this Root type!).   

Should we add all these back-compat switches to the Query subcommand, and then have them accessable to the Root when no subcommand is provided.

stuartpa:
I've added discussion https://github.com/microsoft/go-sqlcmd/discussions/193 to work through how we do this, we can use this PR as a proof point.  

Part of the issue here is we need to make, in this case, the -i switch available on the root command (for back-compat), and on the query subcommand.  What's the best way to wire this up.  Add the -i flag to query.go, and somehow provide a mechanism to also have it made available on the root command (when no sub-command is presented).

We also need to deal with when to show help.  Presuming we make it so the -i command works when no sub-command is added, should we show help for -i when `sqlcmd --help` is run?

Issue #179, opened 2022-12-28
-b command take precedence over :On Error ignore command


JyotikaGargg:
When a user specifies following command :
"sqlcmd on-error.sql -b"

In on-error.sql : 
```
:ONERROR ignore

select @@versionn

GO

select @@version

GO
```

Actual Output : It is exiting with error code.
Expected Output : It should run with only printing error message

shueybubbles:
@JyotikaGargg let's get this fixed then we can send a new release to customers.

PR #181, opened 2023-01-05
on error ignore precedence over -b

JyotikaGargg:
https://github.com/microsoft/go-sqlcmd/issues/179
When a user specifies following command :
"sqlcmd on-error.sql -b"

In on-error.sql :

:ONERROR ignore

select @@versionn

GO

select @@version

GO
Actual Output : It is exiting with error code.
Expected Output : It should run with only printing error message

### DIFF
--- a/pkg/sqlcmd/commands.go
+++ b/pkg/sqlcmd/commands.go
@@ -103,9 +103,9 @@ func newCommands() Commands {
 			isSystem: true,
 		},
 		"ONERROR": {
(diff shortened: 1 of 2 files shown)


shueybubbles:
pls add a test that fails without this fix

shueybubbles (changes requested):
:clock1:

apoorvdeshmukh (commented):
Can you also extend or add a testcase for the issue mentioned in PR?

apoorvdeshmukh (approved):

shueybubbles (approved):
:shipit:

PR #192, opened 2023-01-13
on error name back-compat with odbc

JyotikaGargg:
ONERROR -> replaced with on error for naming connvention matching with ODBC

### DIFF
--- a/pkg/sqlcmd/commands.go
+++ b/pkg/sqlcmd/commands.go
@@ -103,7 +103,7 @@ func newCommands() Commands {
 			isSystem: true,
 		},
 		"ONERROR": {
-			regex:  regexp.MustCompile(`(?im)^[\t ]*?:?ONERROR(?:[ \t]+(.*$)|$)`),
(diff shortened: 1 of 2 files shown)


pkg/sqlcmd/commands.go
```diff
@@ -103,7 +103,7 @@ func newCommands() Commands {
 			isSystem: true,
 		},
 		"ONERROR": {
-			regex:  regexp.MustCompile(`(?im)^[\t ]*?:?ONERROR(?:[ \t]+(.*$)|$)`),
+			regex:  regexp.MustCompile(`(?im)^[\t ]*?:?ON ERROR(?:[ \t]+(.*$)|$)`),
```
apoorvdeshmukh, line 106: Does it also handle lowercase?
Should we have one of the unit tests in lowercase variant of this command?

shueybubbles: The (?im) modifier make the expression case insensitive

See regex101.com for a great interactive tool to play with regex

shueybubbles (approved):
:shipit:

stuartpa (approved):
:shipit:

PR #201, opened 2023-01-25
Fix issues with batch and command parsing

### LINKED ISSUE #197: Parsing error with database name
Hello,

Thank you for addressing the issue with the `__` variable names yesterday.  Our script generated from SQL Server Data Tools is now parsing further with go-sqlcmd, but not completely yet.  There appears to be another gap (and one oddity that I need to think about before complaining about it).

The script contains a section similar to this fairly boiled-down repro:

```
GO
:setvar DatabaseName "MyTestDatabaseGitHub20230120"

GO

IF (DB_ID(N'$(DatabaseName)') IS NOT NULL)
BEGIN
    ALTER DATABASE [$(DatabaseName)]
    SET SINGLE_USER WITH ROLLBACK IMMEDIATE;
    DROP DATABASE [$(DatabaseName)];
END
GO

CREATE DATABASE [$(DatabaseName)]

GO
USE [$(DatabaseName)]
GO
```

When I save the above script to a `.sql` file and run it with go-sqlcmd, using `./sqlcmd -b -i testdb.sql -U SA`, this script returns the following error:

```
Msg 105, Level 15, State 1, Server c6c5144e6044, Line 2
Unclosed quotation mark after the character string ') IS NOT NULL)
BEGIN
    ALTER DATABASE [MyTestDatabaseGitHub20230120]
    SET SINGLE_USER WITH ROLLBACK IMMEDIATE;
    DROP DATABASE [MyTestDatabaseGitHub20230120];
END'.
```

When run with legacy sqlcmd using the same options, the output looks like the below, which is the desired output, and the database is dropped and created.  (note the very first time you run it, it won't output the first four lines because it does not enter the if block):

```
2023-01-20 16:56:50.44 spid52      Setting database option SINGLE_USER to ON for database 'MyTestDatabaseGitHub20230120'.
2023-01-20 16:56:50.82 spid52      Starting up database 'MyTestDatabaseGitHub20230120'.
2023-01-20 16:56:50.87 spid52      Parallel redo is started for database 'MyTestDatabaseGitHub20230120' with worker pool size [1].
2023-01-20 16:56:50.90 spid52      Parallel redo is shutdown for database 'MyTestDatabaseGitHub20230120' with worker pool size [1].
Changed database context to 'MyTestDatabaseGitHub20230120'.
```

Please let me know if I can provide any additional information.  I believe this is some sort of parsing/replacement bug but my Go skills are insufficient to create a failing unit test.

shueybubbles:
- If the first line of a batch was blank and the second line had a variable reference in it, we were calculating the offset to the variable incorrectly. 
- Identifiers surrounded by `[]` were not being properly read
- `exit` command had a buggy regex which matched variants like `exit_something`

Fixes #197 
Fixes #198


nycdotnet (commented):
Is this going to fix other labels that start with commands such as `Rollback_Record:` ?

shueybubbles:
`rollback` isn't a sqlcmd command so I didn't think it was affected.

nycdotnet:
@shueybubbles SQLCMD was throwing errors for me on labels like `Rollback_Record:` and renaming them to something like `zRollback_Record:` fixed it.  Perhaps this is separate - let me try.

nycdotnet:
OK seems to work for me on this version.  👍

nycdotnet (approved):
Not a project approver but these changes appear to fix our issues with v0.11

stuartpa:
Looks good to me (100% code coverage for fix)

stuartpa (approved):
:shipit:

nycdotnet:
Thank you for doing this.  I know you're dealing with some license generation stuff now and the build is failing, but when you get that squared away, doing a release would be very much appreciated.  Thanks for all of your hard work on robustifying go-sqlcmd to work with our legal but gross T-SQL :-)

Issue #175, opened 2022-12-07
Add --attach-db <remote-url> flag to `sqlcmd install mssql`


stuartpa:
SQL Training classrooms want all students to get up and running with a local SQL Server with the class sample database attached in less than a minute (plus any non-cached image download time), so enable:

1. winget | brew | choco | apt install sqlcmd
2. sqlcmd install mssql --attach-db https://github.com/Microsoft/sql-server-samples/releases/download/adventureworks/AdventureWorks2019.bak
3. sqlcmd query "SELECT * FROM [Sales.SalesOrderDetail]"

* sqlcmd would automatically set the default database to be the --attach-database

PR #205, opened 2023-01-30
Round out the sqlcmd install scenario

stuartpa:
This PR rounds out the sqlcmd install scenario based on interviews with users.

It adds `--using https://url.to/db.bak`, e.g. 

`sqlcmd install mssql --using https://aka.ms/AdventureWorksLT.bak`

This downloads and restores a DB as part of the sqlcmd create/install command (in the future we'll allow other file types such as .bacpac)

It also adds `sqlcmd open ads`

The first thing the user wants to do with the sparkly new sql instance is use the UI to query etc.  `sqlcmd open ads` does just that.

It adds `sqlcmd start` and `sqlcmd stop` to start/stop container instances.

Resolves:

https://github.com/microsoft/go-sqlcmd/issues/175

This PR has merged in the following branch (that mocks out dependency on actually bringing up a SQL Server instance for the create/install scenario), which brings unit test execution to around 10 seconds (on a laptop):

  stuartpa/improve-unit-test-verification (https://github.com/microsoft/go-sqlcmd/pull/178)


go.mod
```diff
@@ -13,9 +14,11 @@ require (
 	github.com/microsoft/go-mssqldb v0.20.0
 	github.com/peterh/liner v1.2.2
 	github.com/pkg/errors v0.9.1
+	github.com/sirupsen/logrus v1.9.0
```
shueybubbles, line 17: >github.com/sirupsen/logrus v1.9.0 [](http://example.com/codeflow?start=1&length=33)

run build/build to update the NOTICE file

cmd/modern/main.go
```diff
@@ -95,7 +95,7 @@ func initializeCallback() {
 // nil.  Pass (inject) checkErr into all dependencies (internal helpers etc.) as an
 // errorHandler.
 //
-// To aid debugging issues, if the logging level is > 2 (e.g. -v 3 or -v 4), we
```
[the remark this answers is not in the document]
shueybubbles: --verbosity is good for me

Q - what's the standard for when switches use an = sign?

--verbosity=4
or
--verbosity 4 ?

Some apps seem to have a pattern that the "short" flag doesn't use an = but the long one does

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/microsoft/go-sqlcmd
 go 1.18
 
 require (
+	github.com/alecthomas/assert/v2 v2.1.0
```
shueybubbles, line 6: >alecthomas/assert/v2 v2.1.0 [](http://example.com/codeflow?start=12&length=27)

doesn't stretchr/testify already have asserts?

internal/mssql/mock.go
```diff
@@ -0,0 +1,25 @@
...
+		Connect:           nil,
+		Format:            nil,
+		Query:             "",
+		Cmd:               nil,
+		PrintError:        nil,
+		UnicodeOutputFile: false,
+	}
+}
+
+func (m *MssqlMock) Query(s *sqlcmd.Sqlcmd, text string) {
```
shueybubbles, line 23: This dependency should be defined by an interface instead of 2 types being defined in this package.

Whatever public methods you need for both MssqlMock and MssqlType should be defined by an interface, and the product code will implement the interface using MssqlType and the test code will implement it using whatever type it wants.

mssql.New would take an optional implementation of that interface and use MssqlType if not provided.

If the interface is public the implementation types can probably stay private.

cmd/modern/root/start.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+		Run: c.run,
+	}
+
+	c.Cmd.DefineCommand(options)
+	c.AddFlag(cmdparser.FlagOptions{
+		String:        &c.errorLogEntryToWaitFor,
+		DefaultString: "Recovery is complete",
+		Name:          "errorlog-wait-line",
+		Usage:         "Line in errorlog to wait for before exiting this process",
```
shueybubbles, line 35: this error log reference seems very SQL Server specific.  Why would we expect the user to provide this, when each implementation of "start" should know what its target's behavior is?  And the error log might be in another language.

go.mod
stuartpa: yep, this was a mistake... removed

go.mod
stuartpa: done

cmd/modern/main.go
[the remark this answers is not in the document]
stuartpa: I think we need some short cut, because a typical support scenario is going to be, please run the command with  "-v 4", and "--verbosity 4" doesn't roll off the tongue.

cmd/modern/root/start.go
stuartpa: Yes, and it doesn't actually work either, which is very strange.  i.e. even if we wait here until this entry in the error log, there is still a few seconds before a connect will succeed.

So we can remove this completely from here, because it isn't working.  But I am using the same technique in the `sqlcmd create` section where I might be able to rely on it being US english on first start.  

But I agree with the sentiment, this is hack, because I don't know of a way to actually know SQL is  ready for connections.  

From memory in our test frameworks we wrote a tool called Wait4Sql.exe that listened for a named share memory section? that meant connections could now succeed.

Should I clone that Wait4Sql technique here.  Not sure?

go.mod
stuartpa: also removed the logrus dependency

cmd/modern/root/start.go
shueybubbles: I'd not block at all, by default, and just make Start completely asynchronous. The script author might want to run a bunch of non-SQL commands while SQL is starting up. The open/connect call should be what blocks, for however long a timeout the user has specified.

cmd/modern/root/open/ads_darwin.go
```diff
@@ -0,0 +1,14 @@
+package open
+
+import (
+	"github.com/microsoft/go-sqlcmd/cmd/modern/sqlconfig"
+	"github.com/microsoft/go-sqlcmd/internal/cmdparser"
+)
+
+type Ads struct {
```
shueybubbles, line 8: >Ads [](http://example.com/codeflow?start=5&length=3)

I don't think this acronym has enough public traction yet. 
At minimum it needs a go-doc-compatible comment that spells out that it's for Azure Data Studio. Renaming it AzureDataStudio doesn't seem too verbose, plus it gets the product name a bit more visibility in search.

internal/tools/tool/base_test.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"os"
+	"testing"
+)
+
+func TestName(t *testing.T) {
+	base := Base{name: "test_tool"}
+	if base.Name() != "test_tool" {
+		t.Errorf("expected 'test_tool', but got %v", base.Name())
+	}
```
shueybubbles, line 12: why not use asserts?

internal/tools/tool/base.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (t *Base) ExeFullPath() string {
+	return t.exeName
+}
+
+func (t *Base) SetExeName(exeName string) {
+	t.exeName = exeName
+}
+
+func (t *Base) SetToolDescription(toolYaml Description) {
```
shueybubbles, line 45: >toolYaml [](http://example.com/codeflow?start=34&length=8)

toolYaml --> description ?

internal/tools/tool/base.go
```diff
@@ -0,0 +1,115 @@
...
+	"bytes"
+	"fmt"
+	"os/exec"
+	"runtime"
+	"strings"
+)
+
+type Base struct {
+	name              string
+	isInstalledCalled bool
```
shueybubbles, line 13: >isInstalledCalled bool [](http://example.com/codeflow?start=1&length=22)

you could change `installed` to a `*bool` to avoid a separate flag

internal/tools/tool/ads.go
```diff
@@ -0,0 +1,64 @@
...
+	//   User non-Insiders install
+	//   System non-Insiders install
+	searchLocations := []string{
+		filepath.Join(userProfile, "AppData\\Local\\Programs\\Azure Data Studio - Insiders\\azuredatastudio-insiders.exe"),
+		filepath.Join(programFiles, "Azure Data Studio - Insiders\\azuredatastudio-insiders.exe"),
+		filepath.Join(userProfile, "AppData\\Local\\Programs\\Azure Data Studio\\azuredatastudio.exe"),
+		filepath.Join(programFiles, "Azure Data Studio\\azuredatastudio.exe"),
+	}
+
+	t.Base.SetExeName("azuredatastudio")
```
shueybubbles, line 30: >xeName("azuredatastudio") [](http://example.com/codeflow?start=12&length=25)

in the past, I had problems getting azuredatastudio.exe to properly parse command line parameters and we had to invoke the wrapper `.cmd` in the `bin` folder instead.
Has that situation changed?

cmd/modern/root/open/ads_darwin.go
stuartpa: The name of this type follows the sub-command name, which is `sqlcmd open ads`.  In the internal\tools folder, I have renamed the type to AzureDataStudio.  I will add comments here to explain this.

shueybubbles:
> // Package Output provides a number of methods for logging and handling

is this Output really "Diagnostics" and not meant to be "query output"? 
If it's meant to be the destination for all app output for the modern CLI commands, then the `query` command should probably redirect the `sqlcmd` package output to it.

---
Refers to: internal/output/output.go:4 in 1944095. [](commit_id = 19440959ec580193b3a221ddf01e64b2db9c1f68, deletion_comment = False)

internal/sql/sql.go
```diff
@@ -58,10 +64,15 @@ func Connect(
...
-	s.Query = text
-	s.SetOutput(os.Stdout)
-	s.SetError(os.Stderr)
-	err := s.Run(true, false)
-	checkErr(err)
+func (m *SqlType) Query(text string) {
+	if m.console == nil {
+		m.sqlcmd.Query = text
+		m.sqlcmd.SetOutput(os.Stdout)
+		m.sqlcmd.SetError(os.Stderr)
```
shueybubbles, line 71: See related comment about the `Output` package. Should this command route to the active `Output` instead of standard ?

internal/secret/secret.go
```diff
@@ -47,3 +49,25 @@ func Decode(cipherText string, decryptPassword bool) (plainText string) {
...
+	runes := utf16.Encode([]rune(secret))
+
+	const BOM = '\ufffe' //LE BOM
+	var b [2]byte
+	b[0] = BOM >> 8
+	b[1] = BOM & 255
+	for _, r := range runes {
+		b[1] = byte(r >> 8)
+		b[0] = byte(r & 255)
+		buf.Write(b[:])
```
shueybubbles, line 69: >buf.Write(b[:]) [](http://example.com/codeflow?start=2&length=15)

Isn't this skipping a `Write` of the BOM?

internal/io/file/file.go
```diff
@@ -48,9 +54,29 @@ func Exists(filename string) (exists bool) {
...
+	if err != nil {
+		panic(err)
+	}
+
+	return string(b)
+}
+
+func OpenFile(filename string) *os.File {
+	f, err := os.Create(filename)
+	checkErr(err)
```
shueybubbles, line 68: >checkErr(err) [](http://example.com/codeflow?start=1&length=13)

can the app layer implementations of checkErr similar to how there can be multiple levels of a try/catch for exceptions?
If there's only a global error handler referenced by `checkErr` then any code that needs individualized error handling isn't going to use this wrapper.

internal/credman/credman_windows_test.go
```diff
@@ -0,0 +1,107 @@
...
+		Comment:        "TestComment",
+		LastWritten:    time.Now(),
+		UserName:       "TestUsername",
+		Persist:        PersistSession,
+		CredentialBlob: []byte{0x65, 0x66},
+	}
+
+	err := WriteCredential(credential, CredTypeGeneric)
+	if err != nil {
+		t.Errorf("WriteCredential returned an error: %v", err)
```
shueybubbles, line 20: >Errorf( [](http://example.com/codeflow?start=4&length=7)

this one could be `Fatalf`, unless you are just making sure that if WriteCredential somehow wrote the credential but still returned an error it will get deleted.

internal/credman/types_windows.go
```diff
@@ -0,0 +1,69 @@
...
+	Comment        string
+	LastWritten    time.Time
+	CredentialBlob []byte
+	TargetAlias    string
+	UserName       string
+	Persist        CredentialPersistence
+}
+
+// CredentialPersistence describes one of three persistence modes of a credential.
+// A detailed description of the available modes can be found on
```
shueybubbles, line 26: > on [](http://example.com/codeflow?start=61&length=3)

add link

internal/container/controller.go
```diff
@@ -203,6 +224,84 @@ func (c Controller) ContainerFiles(id string, filespec string) (files []string)
...
+		panic("Must pass in non-empty id")
+	}
+	if src == "" {
+		panic("Must pass in non-empty src")
+	}
+	if dest == "" {
+		panic("Must pass in non-empty dest")
+	}
+
+	cmd := []string{"mkdir", "/var/opt/sql/backup"}
```
shueybubbles, line 238: >"/var/opt/sql/backup" [](http://example.com/codeflow?start=26&length=21)

dest ?

internal/container/controller.go
```diff
@@ -203,6 +224,84 @@ func (c Controller) ContainerFiles(id string, filespec string) (files []string)
...
+	c.runCmdInContainer(id, cmd)
+
+	_, file := filepath.Split(src)
+
+	// Wget the .bak file from the http src, and place it in /var/opt/sql/backup
```
shueybubbles, line 243: > .bak file [](http://example.com/codeflow?start=12&length=10)

this package isn't meant to be SQL backup file specific is it?

internal/container/controller.go
```diff
@@ -203,6 +224,84 @@ func (c Controller) ContainerFiles(id string, filespec string) (files []string)
...
+	cmd = []string{
+		"wget",
+		"-O",
+		"/var/opt/sql/backup/" + file, // not using filepath.Join here, this is in the *nix container. always /
```
shueybubbles, line 247: >"/var/opt/sql/backup/" [](http://example.com/codeflow?start=2&length=22)

dest

cmd/modern/root/open/ads.go
```diff
@@ -0,0 +1,95 @@
...
+		"-r",
+		fmt.Sprintf(
+			"--server=%s", fmt.Sprintf(
+				"%s,%d",
+				localhost,
+				port)),
+	}
+
+	if username != "" {
+		args = append(args, fmt.Sprintf("--user=%s", username))
```
shueybubbles, line 85: >sername)) [](http://example.com/codeflow?start=48&length=9)

ssms passes `--integrated` explicitly for Windows auth. That might be useful if the user has multiple saved connections to the given server and one of those uses integrated auth. Otherwise ADS might ask the user which one to use

cmd/modern/root/open/ads.go
```diff
@@ -0,0 +1,95 @@
...
+	controller := container.NewController()
+	if !controller.ContainerRunning(endpoint.AssetDetails.ContainerDetails.Id) {
+		output.FatalfWithHintExamples([][]string{
+			{"To start the container", "sqlcmd start"},
+		}, "Container is not running")
+	}
+}
+
+// launchAds launches the Azure Data Studio using the specified server and username.
+func (c *Ads) launchAds(localhost string, port int, username string) {
```
shueybubbles, line 73: >localhost  [](http://example.com/codeflow?start=24&length=10)

localhost --> host

shueybubbles, line 73: this should take a database name too

cmd/modern/root/open/ads.go
shueybubbles, line 85: >username [](http://example.com/codeflow?start=47&length=8)

Here's a fun SQL Server behavior  - it allows you to create database and login names that include the `"` character. SSMS escapes those with `\"` when invoking ADS on the command line.

internal/cmdparser/test.go
```diff
@@ -45,16 +54,33 @@ func TestCmd[T PtrAsReceiverWrapper[pointerType], pointerType any](args ...strin
...
-		panic("Only provide one string of args, they will be split on space")
+		panic("Only provide one string of args, they will be split on space/quoted values (with spaces)")
 	} else if len(args) == 1 {
-		c.SetArgsForUnitTesting(strings.Split(args[0], " "))
+		c.SetArgsForUnitTesting(splitStringIntoArgsSlice(args[0]))
 	} else {
 		c.SetArgsForUnitTesting([]string{})
 	}
 	err := c.Command().Execute()
 	return err
```
shueybubbles, line 69: >urn err [](http://example.com/codeflow?start=4&length=7)

should this return the output as well so a test can verify the command did more than not crash?

cmd/modern/root/start.go
stuartpa: Ok, for now I'll remove all this code that is trying to wait until SQL is ready.

Because we can't control all the clients that are going to connect after the start command, I have a feeling this won't be a  good experience.  But we can tackle that later maybe with a Wait4Sql.exe type approach

cmd/modern/root/install/mssql-base.go
```diff
@@ -350,6 +419,106 @@ CHECK_POLICY=OFF`
...
+}
+
+func (c *MssqlBase) downloadImage(imageName string, output *output.Output, controller *container.Controller) {
+	output.Infof("Downloading %v", imageName)
+	err := controller.EnsureImage(imageName)
+	if err != nil || c.unitTesting {
+		output.FatalfErrorWithHints(
+			err,
+			[]string{
+				"Is a container runtime installed on this machine (e.g. Podman or Docker)?" + pal.LineBreak() +
```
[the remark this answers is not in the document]
stuartpa: Yes.  But I don't full understand what the correct semantics of SqlcmdEol are right now.  I don't know if the line break style is based on the OS, or the input/output file format (independent of the OS, but often aligned with the OS.  e.g. sqlcmd running on Windows might be using an input file from Linux, what should EOL be then?

internal/tools/tool/base.go
stuartpa: Fixed

internal/tools/tool/base.go
```diff
@@ -0,0 +1,115 @@
...
+	installed         bool
+	lookPathError     error
+	exeName           string
+	exeFullPath       string
+	toolDescription   Description
```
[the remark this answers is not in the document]
stuartpa: Agree.  Fixed

internal/container/controller.go
stuartpa: /var/opt/sql/backup is the well known folder to place .bak files in. (this is a common pattern in blog posts etc.

internal/container/controller.go
stuartpa: /var/opt/sq/backup is the normal place to place .bak files, used in blog posts etc.

internal/container/controller.go
stuartpa: No, the PR is scoped to enable .bak files to get feedback on whether we are on the right track here.

I've added issue:

[https://github.com/microsoft/go-sqlcmd/issues/210](https://github.com/microsoft/go-sqlcmd/issues/210)

cmd/modern/root/open/ads.go
stuartpa: fixed

cmd/modern/root/open/ads.go
stuartpa: Ouch, good catch.  Fixed

args = append(args, fmt.Sprintf("--user=%s", strings.Replace(username, `"`, `\"`, -1)))

cmd/modern/root/open/ads.go
stuartpa: At the moment it doesn't because the database name is set by the login (via default database), and there isn't a flag on sqlcmd open ads to override this.  So in the interest of scope creep in this PR, I'll leave as is (e.g. it does connect to the user database not master, if a --using is specific, or a --user-database is specified in the sqlcmd create statement.

cmd/modern/main.go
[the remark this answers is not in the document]
stuartpa: Entered issue for this:

[https://github.com/microsoft/go-sqlcmd/issues/211](https://github.com/microsoft/go-sqlcmd/issues/211)

cmd/modern/root/open/ads.go
stuartpa: Done, I've put that in a runtime.GOOS == "windows" check which I presume is correct

internal/credman/types_windows.go
stuartpa: fixed

internal/io/file/file.go
stuartpa: Added issue:

[https://github.com/microsoft/go-sqlcmd/issues/212](https://github.com/microsoft/go-sqlcmd/issues/212)

cmd/modern/root/install/mssql-base.go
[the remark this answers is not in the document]
shueybubbles: sqlcmdeol is meant to be the output line break and it's based only on the OS.
Input can be any kind of line break on any OS; the batch scanner will handle both.

internal/container/controller.go
shueybubbles: sure, but this function has a "dest" parameter. Shouldn't the caller be specifying the var location?

internal/tools/tool/ads.go
stuartpa: I can say it is working with this code, but the current surface area of go-sqlcmd doesn't let much complexity come down to this level at the moment.

internal/tools/tool/base_test.go
stuartpa: Replaced all use of t.Errorf, there was a lot!

Entered:

[https://github.com/microsoft/go-sqlcmd/issues/214](https://github.com/microsoft/go-sqlcmd/issues/214)

To ensure we don't regress the new clean state.

internal/cmdparser/test.go
stuartpa: Good feedback, it should (I was concentrating more on getting high cover coverage, but you are right, we need better verification.  

I think the right abstraction here, is a function with an extra parameter, something TestCmd(args, regExToVerify).

Addedissue for this:

[https://github.com/microsoft/go-sqlcmd/issues/215](https://github.com/microsoft/go-sqlcmd/issues/215)

internal/credman/credman_windows_test.go
stuartpa: This is now a assert.NoErrorf

internal/tools/tool/base.go
stuartpa: Good feedback, fixed

internal/secret/secret.go
stuartpa: Yes, that's why it works!  Removed creation of the BOM, ADS doesn't want it. (auth fails if i write it here)

stuartpa:
> // Package Output provides a number of methods for logging and handling

Yeah, it does go os.Stdout right now:

Set in main.go:

func initializeCallback() {

 // Assigns a new outputter now that we have the outputType and loggingLevel
 // provided to us from the user
 outputter = output.New(
  output.Options{
   StandardWriter: os.Stdout,

I don't fully understand this space right now.  Should this output package always direct all output to the "sqlcmd package output"? 

---
In reply to: [1422933276](https://github.com/microsoft/go-sqlcmd/pull/205#issuecomment-1422933276) [](http://example.com/codeflow?ancestors=1422933276,1422933276)

---
Refers to: internal/output/output.go:4 in 1944095. [](commit_id = 19440959ec580193b3a221ddf01e64b2db9c1f68, deletion_comment = False)

internal/mssql/mock.go
stuartpa: The PR was in draft in iteration 1, I was mid way though implementing... I have a mock impl available in non _test, so the unit tests can call `a sqlcmd install mssql` amd use dummy containers that don't even have SQL Server in them (the Hello World docker container), and then use this mock imple to fake out the Connect/SQL (that does the sa disable/create user account etc.), this keeps the unit testing runtime down to just a few seconds for all tests.

internal/sql/sql.go
stuartpa: Right now because we haven't wired up all the Kong switches, there is no way to set different output/error targets, so I set it to these defaults now,  But when we start wiring up all the back-compat switches to the query subcommand these will have to change.

stuartpa:
> // Package Output provides a number of methods for logging and handling

I think the code is right for now, because we haven't wired up the back-compat switches to the modern cli sub-commands.  

---
In reply to: [1426724756](https://github.com/microsoft/go-sqlcmd/pull/205#issuecomment-1426724756) [](http://example.com/codeflow?ancestors=1426724756,1422933276,1422933276)

---
Refers to: internal/output/output.go:4 in 1944095. [](commit_id = 19440959ec580193b3a221ddf01e64b2db9c1f68, deletion_comment = False)

PR #207, opened 2023-02-06
Add localization support for go-sqlcmd

apoorvdeshmukh:
This commit adds localization framework and support for languages 
en-us, de-de, fr-fr, es-es, pt-br, ko-kr, zh-cn, zh-tw, it-it, ja-jp, ru-ru. 
Localized text added for de-de, fr-fr and es-es. Subsequqnt commits will
 add localized text for remaining languages. Translations have been taken
 from ODBC sqlcmd.rll.lcl for corresponding language.


internal/localizer/localizer.go
```diff
@@ -0,0 +1,109 @@
+package localizer
+
+import (
+	// Import the internal/translations so that it's init() function
```
shueybubbles, line 4: >it's [](http://example.com/codeflow?start=45&length=4)

its for possessive

internal/localizer/localizer.go
```diff
@@ -0,0 +1,109 @@
...
+		ID:      "pt-br",
+		printer: message.NewPrinter(language.MustParse("pt-BR")),
+	},
+	{
+		ID:      "ru-ru",
+		printer: message.NewPrinter(language.MustParse("ru-RU")),
+	},
+	{
+		ID:      "es-es",
+		printer: message.NewPrinter(language.MustParse("es-ES")),
```
shueybubbles, line 71: >NewPrinter(language.MustParse("es-ES")) [](http://example.com/codeflow?start=19&length=39)

is NewPrinter a func or a struct? If it's a struct, can we change  Localizer.printer to a func so it doesn't allocate memory until it's called? We should avoid creating a bunch of objects for all the languages when we're only going to use one of them

internal/localizer/localizer.go
```diff
@@ -0,0 +1,109 @@
...
+	},
+}
+
+// The Get() function accepts a locale ID and returns the corresponding
+// Localizer for that locale. If the locale ID is not supported then
+// this returns `false` as the second return value.
+func Get(id string) Localizer {
+	for _, locale := range locales {
+		if id == locale.ID {
```
shueybubbles, line 80: >id == locale.ID  [](http://example.com/codeflow?start=5&length=16)

use EqualFold() for case insensitivity

internal/localizer/localizer.go
```diff
@@ -0,0 +1,109 @@
...
+	{
+		ID:      "fr-fr",
+		printer: message.NewPrinter(language.MustParse("fr-FR")),
+	},
+	{
+		ID:      "en-us",
+		printer: message.NewPrinter(language.MustParse("en-US")),
+	},
+	{
+		ID:      "zh-cn",
```
shueybubbles, line 42: >cn", [](http://example.com/codeflow?start=15&length=4)

the modern lang ids are zh-hans and zh-hant

internal/translations/locales/en-US/out.gotext.json
```diff
@@ -0,0 +1,277 @@
+{
+    "language": "en-US",
```
shueybubbles, line 2: >"language": "en-US", [](http://example.com/codeflow?start=4&length=20)

Is this statement correct:

Eventually,  there will also be a .lci file  here. We will have a pipeline that directs the Microsoft localization tools to generate the translations in the SQL server language sets from this english json file, and that lci file will inform the loc tools to only localize the "language" and "translation" fields in the output json.

We will submit the output of that pipeline to the main branch in the appropriate internal/translations/locales folders.

That lci file would also have some rules to help the translators avoid inappropriate argument reordering and to identify some non-translatable terms in some strings.

internal/translations/locales/es-ES/out.gotext.json
```diff
@@ -0,0 +1,245 @@
+{
+    "language": "es-ES",
+    "messages": [
```
shueybubbles, line 3: pls remind us why we have 2 json files in each folder.
Won't the localization process only need the en-us file for input?

pkg/sqlcmd/errors.go
```diff
@@ -5,18 +5,19 @@ package sqlcmd
...
 	"errors"
-	"fmt"
 	"strings"
+
+	"github.com/microsoft/go-sqlcmd/internal/localizer"
 )
 
 // ErrorPrefix is the prefix for all sqlcmd-generated errors
-const ErrorPrefix = "Sqlcmd: Error: "
+var ErrorPrefix string = localizer.Sprintf("Sqlcmd: Error: ")
```
shueybubbles, line 14: >f("Sqlcmd: Error:  [](http://example.com/codeflow?start=41&length=18)

does the "Sqlcmd" name get localized  in ODBC? If not, is it due to having a comment in the .lci file or does ODBC not have the name as a string resource?

pkg/sqlcmd/errors.go
```diff
@@ -5,18 +5,19 @@ package sqlcmd
...
 
 // WarningPrefix is the prefix for all sqlcmd-generated warnings
-const WarningPrefix = "Sqlcmd: Warning: "
+var WarningPrefix string = localizer.Sprintf("Sqlcmd: Warning: ")
 
 // Common Sqlcmd error messages
-const ErrCmdDisabled = "ED and !!<command> commands, startup script, and environment variables are disabled"
+var ErrCmdDisabled string = localizer.Sprintf("ED and !!<command> commands, startup script, and enviroment variables are disabled")
```
shueybubbles, line 20: >enviroment [](http://example.com/codeflow?start=96&length=10)

environment

pkg/sqlcmd/errors.go
```diff
@@ -73,23 +74,24 @@ func (e *VariableError) IsSqlcmdErr() bool {
 func ReadOnlyVariable(variable string) *VariableError {
 	return &VariableError{
 		Variable:      variable,
-		MessageFormat: "The scripting variable: '%s' is read-only",
+		MessageFormat: localizer.Sprintf("The scripting variable: '%s' is read-only", variable),
```
shueybubbles, line 77: >MessageFormat [](http://example.com/codeflow?start=2&length=13)

technically this is a breaking change but I doubt anything external is depending on the current behavior.
It's probably worth having a note in the merge commit message though.

shueybubbles (changes requested):
:clock1:

internal/localizer/localizer.go
apoorvdeshmukh: Fixed it

internal/localizer/localizer.go
apoorvdeshmukh: I agree. No every language printer was needed to be initialized. I have refactored this code to instantiate only one language printer instance.

internal/localizer/localizer.go
apoorvdeshmukh: This function has been removed

internal/localizer/localizer.go
apoorvdeshmukh: Added modern lang ids as suggested.

internal/translations/catalog.go
```diff
@@ -0,0 +1,284 @@
...
+		"es_ES": &dictionary{index: es_ESIndex, data: es_ESData},
+		"fr_FR": &dictionary{index: fr_FRIndex, data: fr_FRData},
+		"it_IT": &dictionary{index: it_ITIndex, data: it_ITData},
+		"ja_JP": &dictionary{index: ja_JPIndex, data: ja_JPData},
+		"ko_KR": &dictionary{index: ko_KRIndex, data: ko_KRData},
+		"pt_BR": &dictionary{index: pt_BRIndex, data: pt_BRData},
+		"ru_RU": &dictionary{index: ru_RUIndex, data: ru_RUData},
+		"zh_CN": &dictionary{index: zh_CNIndex, data: zh_CNData},
+		"zh_TW": &dictionary{index: zh_TWIndex, data: zh_TWData},
+	}
```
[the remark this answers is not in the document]
apoorvdeshmukh: Unfortunately, we have no control over this file as this is auto generated by go generate command. 
I'll dig more later to see if there are any options to influence the code in this file.

internal/translations/locales/en-US/out.gotext.json
apoorvdeshmukh: This is correct. 
In short, we  provide out.gotext.json file as input 
1. Run LsBuild command which takes this json, lci (file created using LCXAdmin to specify what needs to be localized) and lss (parser related settings file, optional) as input and generates .lcg file.
2. The lcg file is what is provided to translators, and they save it as .lcl after translations are added.
3. Run LsBuild command per language which takes lcl file as input and generates localized '.json' file for that language at specified directory.

pkg/sqlcmd/errors.go
apoorvdeshmukh: 'Sqlcmd' name is not localized in ODBC and it does appear as string resource in lcl file. I have not checked the .lci file but I think somewhere it must have been specified that 'Sqlcmd' as a name should not be localized because in all languages it appears as 'Sqlcmd' only in error messages.

pkg/sqlcmd/errors.go
apoorvdeshmukh: Yeah, I tried multiple things to keep the previous approach but then the scanner failed to recognize what '%s' should be replaced with and couldn't replace %s with anything when actual localized error message was produced.

pkg/sqlcmd/errors.go
apoorvdeshmukh: Corrected it. I had copied error messages from ODBC sqlcmd.rll.lcl to our code to ensure that we have exact same error message and ended up copying the typo as well.

apoorvdeshmukh:
Are there any pending comments or suggestions?

stuartpa (approved):
:shipit:

shueybubbles:
pls Resolve issues as you fix them. Every comment still shows as "Active"

---
In reply to: [1427551524](https://github.com/microsoft/go-sqlcmd/pull/207#issuecomment-1427551524) [](http://example.com/codeflow?ancestors=1427551524,1427551524)

internal/localizer/localizer.go
```diff
@@ -0,0 +1,56 @@
...
+	"golang.org/x/text/message"
+)
+
+var Translator *message.Printer
+
+var supportedLanguages = map[string]string{
+	"de-de":      "de-DE",
+	"fr-fr":      "fr-FR",
+	"en-us":      "en-US",
+	"zh-hans-cn": "zh-CN",
```
shueybubbles, line 24: >zh-hans-cn [](http://example.com/codeflow?start=2&length=10)

I've never seen "zh-hans-cn" used, just "zh-hans"

shueybubbles (approved):
:shipit:

internal/localizer/localizer.go
apoorvdeshmukh: The bcp47 language tag expects the second half as the region so I was not sure what we should have there. 
Since we are anyway internally mapping it to zh-CN and zh-TW, I have corrected this to be zh-hans and zh-hant for user to specify.

PR #208, opened 2023-02-07
Kong to Cobra Migration 

JyotikaGargg:
Kong To Cobra TryOuts :

1. I replaced the code in Execute function in sqlcmd.go with cobra basic command implementation which calls the same function inside its "run" as legacy sqlcmd .
2.  Added InputFiles flags and cobra execute function is able to read them and parse them and got the following output :
In launch.json :
`{
            "name" : "Run query and exit",
            "type" : "go",
            "request": "launch",
            "mode" : "auto",
            "program": "${workspaceFolder}/cmd/modern",
            //"args":["testCommand","GO"]
            // "args" : ["select @@version","GO"],
            //"args":["-i","..\\sqlcmd\\testdata\\select100.sql"] 
            "args":["--Version"]           
        },`
File contains : 
`select 100 
GO`

Output :

`(1 row affected)
GO         
-----------
        100

(1 row affected)`


shueybubbles:
I am not sure how you would get `nil` for `s.Query" because `Query` is a `string`, not a `*string` so its default value is "" and you can't assign a nil to it.

shueybubbles:
> 	LoginTimeout                int               `short:"l" default:"-1" help:"Specifies the number of seconds before a sqlcmd login to the go-mssqldb driver times out when you try to connect to a server. This option sets the sqlcmd scripting variable SQLCMDLOGINTIMEOUT. The default value is 30. 0 means infinite."`

I think you'll need to assign the default values manually now that kong isn't doing it. 
Otherwise the `args` is going to have a bunch of empty strings and zeroes which are invalid values in some cases.

---
Refers to: cmd/sqlcmd/sqlcmd.go:44 in 510ca91. [](commit_id = 510ca9148df77ce64fa2ba6424282c5008d2c876, deletion_comment = False)

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +117,109 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+			fmt.Printf("after execute Function")
+		},
 	}
-	if args.Help {
-		_ = ctx.PrintUsage(false)
-		os.Exit(0)
+
+	//var version string
+
+	rootCmd.PersistentFlags().BoolVarP(&args.Version, "version", "v", false, "Print version information and exit")
```
shueybubbles, line 152: > "v" [](http://example.com/codeflow?start=61&length=4)

there's no short flag for this parameter

JyotikaGargg: Removed short flag.

JyotikaGargg:
> I am not sure how you would get `nil` for `s.Query` because `Query` is a `string`, not a `*string` so its default value is "" and you can't assign a nil to it.

Got this resolved . It was "" though ,  I was passing some wrong values.

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +117,50 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
-	setVars(vars, &args)
 
-	// so far sqlcmd prints all the errors itself so ignore it
-	exitCode, _ := run(vars, &args)
-	os.Exit(exitCode)
+	var inputfiles []string
+	rootCmd.Flags().StringArrayVarP(&args.InputFile, "i", "i", inputfiles, "input file")
+	rootCmd.PersistentFlags().BoolVarP(&args.Version, "Version", "", false, "Print version information and exit")
+	rootCmd.PersistentFlags().BoolVarP(&args.Help, "help", "h", false, "Print this help message and exit")
+	rootCmd.PersistentFlags().BoolVarP(&args.DisableCmdAndWarn, "disable-cmd-and-warn", "", false, "Description of DisableCmdAndWarn")
```
shueybubbles, line 154: >"" [](http://example.com/codeflow?start=85&length=2)

"X"

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +117,50 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+		},
 	}
-	vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
-	setVars(vars, &args)
 
-	// so far sqlcmd prints all the errors itself so ignore it
-	exitCode, _ := run(vars, &args)
-	os.Exit(exitCode)
+	var inputfiles []string
+	rootCmd.Flags().StringArrayVarP(&args.InputFile, "i", "i", inputfiles, "input file")
```
shueybubbles, line 151: >StringArrayVar [](http://example.com/codeflow?start=17&length=14)

This should probably be StringSliceVarP so cobra handles

```
-i file1,file2 -i file3
```
correctly

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+			if args.Help {
+				fmt.Print(cmd.UsageString())
+				os.Exit(0)
+			}
+
+			exitCode, err := run(vars, &args)
+			if err != nil {
+				fmt.Println(err)
+				os.Exit(1)
+			}
```
shueybubbles, line 143: we don't need to print  the error from run because run already prints relevant errors.

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+
+			fmt.Printf("after exit exitCode: %d\n", exitCode)
+			//os.Exit(exitCode)
+			fmt.Printf("after execute Function")
```
shueybubbles, line 147: delete the extra debugging prints

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+				fmt.Printf("%v\n", version)
+				os.Exit(0)
+			}
+
+			if args.Help {
+				fmt.Print(cmd.UsageString())
+				os.Exit(0)
+			}
+
+			exitCode, err := run(vars, &args)
```
shueybubbles, line 139: >exitCode [](http://example.com/codeflow?start=3&length=8)

pass exitCode to the os.Exit function

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+		os.Exit(1)
 	}
-	vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
-	setVars(vars, &args)
+}
+
+func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
+	var inputfiles []string
+	rootCmd.Flags().StringArrayVarP(&args.InputFile, "i", "i", inputfiles, "input file")
+	rootCmd.PersistentFlags().BoolVarP(&args.Version, "Version", "", false, "Print version information and exit")
```
shueybubbles, line 159: > "Print version information and exit") [](http://example.com/codeflow?start=72&length=38)

wrap the descriptions to make them localizable

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	rootCmd.PersistentFlags().StringVar(&args.AuthenticationMethod, "AuthenticationMethod", "", "Specifies the SQL authentication method to use to connect to Azure SQL Database. One of:ActiveDirectoryDefault,ActiveDirectoryIntegrated,ActiveDirectoryPassword,ActiveDirectoryInteractive,ActiveDirectoryManagedIdentity,ActiveDirectoryServicePrincipal,SqlPassword")
+	rootCmd.PersistentFlags().BoolVarP(&args.UseAad, "UseAad", "G", false, "Tells sqlcmd to use Active Directory authentication. If no user name is provided, authentication method ActiveDirectoryDefault is used. If a password is provided, ActiveDirectoryPassword is used. Otherwise ActiveDirectoryInteractive is used.")
+	rootCmd.PersistentFlags().BoolVarP(&args.DisableVariableSubstitution, "DisableVariableSubstitution", "x", false, "Causes sqlcmd to ignore scripting variables. This parameter is useful when a script contains many INSERT statements that may contain strings that have the same format as regular variables, such as $(variable_name).")
```
shueybubbles, line 168: >"DisableVariableSubstitution [](http://example.com/codeflow?start=71&length=28)

does Cobra automatically convert these strings to switches like --disable-variable-substitution?

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	rootCmd.Flags().StringVarP(&args.Format, "Format", "F", "horiz", "Specifies the formatting for results.")
+	rootCmd.Flags().IntVarP(&args.ErrorsToStderr, "ErrorsToStderr", "r", -1, "Controls which error messages are sent to stdout. Messages that have severity level greater than or equal to this level are sent.")
+
+	//Adding a validator for checking the enum flags
+	rootCmd.Flags().SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
+		switch name {
+		case "ApplicationIntent":
+			value := getFlagValueByName(f, name)
+			switch value {
+			case "default", "ReadOnly":
```
shueybubbles, line 186: >se "default", "ReadOnly": [](http://example.com/codeflow?start=5&length=25)

make sure we preserve any case insensitivity from the odbc sqlcmd

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+				return pflag.NormalizedName(name)
+			default:
+				fmt.Fprintf(os.Stderr, "Invalid value for %s flag. Allowed values are default and ReadOnly.\n", name)
```
shueybubbles, line 189: >Invalid value for %s flag. Allowed values are default and ReadOnly.\n", name) [](http://example.com/codeflow?start=28&length=77)

Use the same error messages that the ODBC sqlcmd uses

eg

Sqlcmd: Option -K: Unexpected argument: foobar

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	var inputfiles []string
+	rootCmd.Flags().StringArrayVarP(&args.InputFile, "i", "i", inputfiles, "input file")
+	rootCmd.PersistentFlags().BoolVarP(&args.Version, "Version", "", false, "Print version information and exit")
+	rootCmd.PersistentFlags().StringVarP(&args.BatchTerminator, "Batch Terminator", "c", "GO", "Specifies the batch terminator. The default value is GO")
+	rootCmd.PersistentFlags().StringVarP(&args.UserName, "User Name", "U", "", "Specifies the batch terminator. The default value is GO")
+	rootCmd.PersistentFlags().StringVarP(&args.InitialQuery, "InitialQuery", "q", "", "Executes a query when sqlcmd starts, but does not exit sqlcmd when the query has finished running. Multiple-semicolon-delimited queries can be executed.")
+	rootCmd.PersistentFlags().StringVarP(&args.Query, "Query", "Q", "", "Executes a query when sqlcmd starts and then immediately exits sqlcmd. Multiple-semicolon-delimited queries can be executed.")
+	rootCmd.PersistentFlags().StringVarP(&args.Server, "Server", "S", "", "[tcp:]server[\\instance_name][,port]Specifies the instance of SQL Server to which to connect. It sets the sqlcmd scripting variable SQLCMDSERVER.")
```
shueybubbles, line 164: >p:]s [](http://example.com/codeflow?start=75&length=4)

we support lpc and np now too

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	rootCmd.Flags().StringVarP(&args.ColumnSeparator, "ColumnSeparator", "s", "", "Specifies the column separator character. Sets the SQLCMDCOLSEP variable.")
+	rootCmd.Flags().BoolVarP(&args.TrimSpaces, "TrimSpaces", "W", false, "Remove trailing spaces from a column.")
+	rootCmd.Flags().BoolVarP(&args.MultiSubnetFailover, "MultiSubnetFailover", "M", false, "Provided for backward compatibility. Sqlcmd always optimizes detection of the active replica of a SQL Failover Cluster.")
+
+	rootCmd.Flags().StringVarP(&args.Password, "Password", "P", "", "Obsolete. The initial passwords must be set using the SQLCMDPASSWORD environment variable or entered at the password prompt.")
+	rootCmd.Flags().BoolVarP(&args.Help, "Help", "?", false, "Show syntax summary.")
+
+	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "ErrorSeverityLevel", "V", 0, "Controls the severity level that is used to set the ERRORLEVEL variable on exit.")
+	//rootCmd.PersistentFlags().IntVarP(args.ScreenWidth, "ScreenWidth", "w", 0, "Specifies the screen width for output. Sets the SQLCMDCOLWIDTH variable.")
```
shueybubbles, line 239: >//rootCmd.PersistentFlags().IntVarP(args.S [](http://example.com/codeflow?start=1&length=42)

you can use Flags().Var() and provide a custom type for special cases.

cmd/sqlcmd/sqlcmd.go
shueybubbles, line 159: >PersistentFlags [](http://example.com/codeflow?start=9&length=15)

since we don't have any child commands, can't all these just be Flags() instead of PersistentFlags() ?

shueybubbles:
please resolve comments for fixed issues when you push changes so it's clear to reviewers whether you've fixed something from an earlier iteration comment.

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,140 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+}
+
+func getFlagValueByName(flagSet *pflag.FlagSet, name string) string {
+	var value string
+	flagSet.VisitAll(func(f *pflag.Flag) {
+		if f.Name == name {
+			value = f.Value.String()
+		}
+	})
+	return value
```
shueybubbles, line 249: can't this return once it finds the flag instead of after VisitAll returns?

shueybubbles (changes requested):
:clock1:

JyotikaGargg:
> I am not sure how you would get `nil` for `s.Query` because `Query` is a `string`, not a `*string` so its default value is "" and you can't assign a nil to it.

Handled.

cmd/sqlcmd/sqlcmd.go
JyotikaGargg: No , it converts the flag names to switches in the format '--flag-name'.
So here it would be '--DisableVariableSubstitution'. But we could override and provide custom name as the second parameter.
So i will override Name parameter by comparing with ODBC

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,137 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+				os.Exit(0)
+			}
+
+			if args.Help {
+				fmt.Print(cmd.UsageString())
+				os.Exit(0)
+			}
+
+			exitCode, err := run(vars, &args)
+			if err != nil {
```
shueybubbles, line 140: >if err != nil { [](http://example.com/codeflow?start=3&length=15)

always pass exitCode to os.Exit regardless of err value

cmd/sqlcmd/sqlcmd.go
shueybubbles: ODBC sqlcmd doesn't have any long names, just short flags.
Let's preserve the long names that the Kong implementation uses since it's already being used.

cmd/sqlcmd/sqlcmd.go
```diff
@@ -117,25 +118,137 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+			value = f.Value.String()
+			return
+		}
+	})
+	return value
 }
 
 // setVars initializes scripting variables from command line arguments
 func setVars(vars *sqlcmd.Variables, args *SQLCmdArguments) {
+	fmt.Printf("in setvar function")
```
shueybubbles, line 251: >fmt.Printf("in setvar function") [](http://example.com/codeflow?start=1&length=32)

delete

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,21 +120,132 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+}
+
+func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
+	var inputfiles []string
+	rootCmd.Flags().StringSliceVarP(&args.InputFile, "input-file", "i", inputfiles, "Identifies one or more files that contain batches of SQL statements. If one or more files do not exist, sqlcmd will exit. Mutually exclusive with -Q/-q.")
+	rootCmd.Flags().BoolVarP(&args.Version, "Version", "", false, "Print version information and exit")
+	rootCmd.Flags().StringVarP(&args.BatchTerminator, "batch-terminator", "c", "GO", "Specifies the batch terminator. The default value is GO")
+	rootCmd.Flags().StringVarP(&args.UserName, "user-name", "U", "", "The login name or contained database user name.  For contained database users, you must provide the database name option")
+	rootCmd.Flags().StringVarP(&args.InitialQuery, "initial-query", "q", "", "Executes a query when sqlcmd starts, but does not exit sqlcmd when the query has finished running. Multiple-semicolon-delimited queries can be executed.")
+	rootCmd.Flags().StringVarP(&args.Query, "Query", "Q", "", "Executes a query when sqlcmd starts and then immediately exits sqlcmd. Multiple-semicolon-delimited queries can be executed.")
```
shueybubbles, line 160: >Query [](http://example.com/codeflow?start=42&length=5)

query

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,21 +120,132 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	var variables map[string]string
+	rootCmd.Flags().StringToStringVarP(&args.Variables, "Variables", "v", variables, "Creates a sqlcmd scripting variable that can be used in a sqlcmd script. Enclose the value in quotation marks if the value contains spaces. You can specify multiple var=values values. If there are errors in any of the values specified, sqlcmd generates an error message and then exits")
+	rootCmd.Flags().IntVarP(&args.PacketSize, "PacketSize", "a", 0, "Requests a packet of a different size. This option sets the sqlcmd scripting variable SQLCMDPACKETSIZE. packet_size must be a value between 512 and 32767. The default = 4096. A larger packet size can enhance performance for execution of scripts that have lots of SQL statements between GO commands. You can request a larger packet size. However, if the request is denied, sqlcmd uses the server default for packet size.")
```
shueybubbles, line 168: >PacketSize [](http://example.com/codeflow?start=44&length=10)

packet-size

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,21 +120,132 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+
+	rootCmd.Flags().StringVarP(&args.ApplicationIntent, "application-intent", "K", "default", "Declares the application workload type when connecting to a server. The only currently supported value is ReadOnly. If -K is not specified, the sqlcmd utility will not support connectivity to a secondary replica in an Always On availability group")
+	rootCmd.Flags().StringVarP(&args.EncryptConnection, "encrypt-connection", "N", "default", "This switch is used by the client to request an encrypted connection.")
+	rootCmd.Flags().StringVarP(&args.Format, "Format", "F", "horiz", "Specifies the formatting for results.")
+	rootCmd.Flags().IntVarP(&args.ErrorsToStderr, "errors-to-stderr", "r", -1, "Controls which error messages are sent to stdout. Messages that have severity level greater than or equal to this level are sent.")
+
+	//Adding a validator for checking the enum flags
+	rootCmd.Flags().SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
```
shueybubbles, line 178: >rootCmd.Flags().SetNormalizeFunc [](http://example.com/codeflow?start=1&length=32)

move this call somewhere so it doesn't break up the overall list of flags

shueybubbles:
note there's also Kong usage in the `:connect` command to address https://github.com/microsoft/go-sqlcmd/blob/978080120d88f55eae0330af371d617d5f11fba7/pkg/sqlcmd/commands.go#L408

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,172 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
-	if args.Version {
-		ctx.Printf("%v\n", version)
-		fmt.Println(localizer.Sprintf("Legal docs and information: aka.ms/SqlcmdLegal"))
-		fmt.Println(localizer.Sprintf("Third party notices: aka.ms/SqlcmdNotices"))
-		os.Exit(0)
+	// Cobra Tryout #2 : This has version and help flags. But they are also getting stucked inside run function without error
+	rootCmd := &cobra.Command{
+		Use:   "testCommand",
+		Short: "A brief description of your command",
+		PersistentPreRunE: func(cmd *cobra.Command, argss []string) error {
```
shueybubbles, line 127: >PersistentPreRunE [](http://example.com/codeflow?start=2&length=17)

why do we have both PersistentPreRunE and PreRunE?

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,172 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
-	// so far sqlcmd prints all the errors itself so ignore it
-	exitCode, _ := run(vars, &args)
-	os.Exit(exitCode)
+	if err := rootCmd.Execute(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func normalizeWithError(name string, err error) (pflag.NormalizedName, error) {
+	if err != nil {
```
shueybubbles, line 167: >f err != nil { [](http://example.com/codeflow?start=2&length=14)

don't you need to check the flag name too?

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,172 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	rootCmd.Flags().StringVarP(&args.ColumnSeparator, "column-separator", "s", "", "Specifies the column separator character. Sets the SQLCMDCOLSEP variable.")
+	rootCmd.Flags().BoolVarP(&args.TrimSpaces, "trim-spaces", "W", false, "Remove trailing spaces from a column.")
+	rootCmd.Flags().BoolVarP(&args.MultiSubnetFailover, "multi-subnet-failover", "M", false, "Provided for backward compatibility. Sqlcmd always optimizes detection of the active replica of a SQL Failover Cluster.")
+
+	rootCmd.Flags().StringVarP(&args.Password, "password", "P", "", "Obsolete. The initial passwords must be set using the SQLCMDPASSWORD environment variable or entered at the password prompt.")
+
+	// Using PersistentFlags() for ErrorSeverityLevel due to data type uint8 , which is not supported in Flags()
+	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "error-severity-level", "V", 0, "Controls the severity level that is used to set the ERRORLEVEL variable on exit.")
+	// ScreenWidth is *int type , need to find corresponding Flag Type in PFlag
```
shueybubbles, line 218: >e , need to find corresponding Flag Type in PFlag [](http://example.com/codeflow?start=27&length=49)

i think you'd use the Lookup function to decide if ScreenWidth will be set to a non-null value

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,172 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	//Adding a validator for checking the enum flags
+	rootCmd.Flags().SetNormalizeFunc(func(f *pflag.FlagSet, name string) pflag.NormalizedName {
+		switch name {
+		case "application-intent":
+			value := strings.ToLower(getFlagValueByName(f, name))
+			switch value {
+			case "default", "readonly":
+				return pflag.NormalizedName(name)
+			default:
+				_, err := normalizeWithError(value, fmt.Errorf("--application-intent must be one of \"default\",\"readonly\",\"vert\",\"vertical\" but got \"%s\"", value))
```
shueybubbles, line 232: >"--application-intent mu [](http://example.com/codeflow?start=51&length=24)

this error has nothing to do with application-intent

shueybubbles, line 232: >\"vert\",\"vertical\"  [](http://example.com/codeflow?start=113&length=22)

remove "vert" and "vertical"

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,172 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	})
+
+}
+
+func getFlagValueByName(flagSet *pflag.FlagSet, name string) string {
+	var value string
+	flagSet.VisitAll(func(f *pflag.Flag) {
+		if f.Name == name {
+			value = f.Value.String()
+			return
```
shueybubbles, line 285: >		return [](http://example.com/codeflow?start=1&length=8)

return value

cmd/sqlcmd/sqlcmd_test.go
```diff
@@ -90,15 +74,23 @@ func TestValidCommandLineToArgsConversion(t *testing.T) {
...
+			Long:  "A long description of my command",
+			Run: func(cmd *cobra.Command, argss []string) {
+				// Command logic goes here
+				vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
+				setVars(vars, &args)
+			},
+		}
+		setFlags(cmd, arguments)
+		cmd.SetArgs(test.commandLine)
+		err := cmd.Execute()
```
shueybubbles, line 93: >	err := cmd.Execute() [](http://example.com/codeflow?start=1&length=21)

we can't unit test Execute because it is going to try to do everything.
Instead make sure the function that populates SQLCmdArguments is individually callable and unit test that function

cmd/sqlcmd/sqlcmd_test.go
```diff
@@ -90,15 +74,23 @@ func TestValidCommandLineToArgsConversion(t *testing.T) {
...
 		arguments := &SQLCmdArguments{}
-		parser := newKong(t, arguments)
-		_, err := parser.Parse(test.commandLine)
+		cmd := &cobra.Command{
+			Use:   "testCommand",
+			Short: "A brief description of my command",
+			Long:  "A long description of my command",
+			Run: func(cmd *cobra.Command, argss []string) {
+				// Command logic goes here
+				vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn)
```
shueybubbles, line 87: >vars := sqlcmd.InitializeVariables(!args.DisableCmdAndWarn) [](http://example.com/codeflow?start=4&length=59)

the kong version of this test never called InitializeVariables or setVars. 
The test is just meant to validate that the command line parsing populates the struct correctly.

cmd/sqlcmd/sqlcmd.go
JyotikaGargg: Its by mistake while testing each . Removing PreRunE.

cmd/sqlcmd/sqlcmd.go
JyotikaGargg: > 
This return exit from loop and then func return value.
I am not able to access http://example.com/codeflow?start=1&length=8

pkg/sqlcmd/commands.go
```diff
@@ -397,44 +397,57 @@ type connectData struct {
...
+			return nil
+		},
 	}
-	connect.AuthenticationMethod = arguments.AuthenticationMethod
-	// If no user name is provided we switch to integrated auth
-	_ = s.ConnectDb(&connect, s.lineIo == nil)
-	// ConnectDb prints connection errors already, and failure to connect is not fatal even with -b option
-	return nil
+
+	cmd.Flags().StringP("server", "", "", "Specifies the server to connect to")
```
shueybubbles, line 443: >StringP("server", " [](http://example.com/codeflow?start=13&length=19)

isn't "server" going to be the positional argument? Does it need to also have a Flags entry?

pkg/sqlcmd/commands.go
```diff
@@ -397,44 +397,57 @@ type connectData struct {
 }
 
 func connectCommand(s *Sqlcmd, args []string, line uint) error {
+	cmd := &cobra.Command{
+		Use:   "connect",
+		Short: "Connect to a database",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
JyotikaGargg: The RunE method, is used when the command action may return an error but Run only prints but not return anything. Here for connect command i used RunE , because connecting to the database may fail and we need to be able to handle that error

pkg/sqlcmd/commands.go
JyotikaGargg: Changed the code .

cmd/sqlcmd/sqlcmd.go
JyotikaGargg: This is going to be taken care in Apoorv's PR

.vscode/launch.json
```diff
@@ -23,8 +23,13 @@
...
             "mode" : "auto",
+<<<<<<< HEAD
+            "program": "${workspaceFolder}/cmd/sqlcmd",
+            "args" : ["-Q", "EXIT(select 100 as Count)"],      
+=======
             "program": "${workspaceFolder}/cmd/modern",
             "args" : ["-Q", "EXIT(select net_transport from sys.dm_exec_connections)"],
+>>>>>>> 5c47bd4fb4d8efcefbeb5af3949ef107ff4e0f4b
         },
         {
```
shueybubbles, line 34: fix merge

pkg/sqlcmd/commands.go
```diff
@@ -387,41 +387,31 @@ func listCommand(s *Sqlcmd, args []string, line uint) (err error) {
...
-		return InvalidCommandError("CONNECT", line)
-	}
-	arguments := &connectData{}
-	parser, err := kong.New(arguments)
-	if err != nil {
-		return InvalidCommandError("CONNECT", line)
-	}
+	// Parse flags
+	flags := flag.NewFlagSet("connect", flag.ContinueOnError)
+	server := flags.String("S", "", "server name")
```
shueybubbles, line 396: :connect doesn't have an S flag
server name is just a positional argument

shueybubbles: I believe you would get it from Flags.Arg(0)

cmd/sqlcmd/testdata/select100.sql
```diff
@@ -1 +1,2 @@
 select 100
+GO
```
shueybubbles, line 2: why do we need this GO?

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,170 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+	rootCmd.Flags().BoolVarP(&args.TrimSpaces, "trim-spaces", "W", false, "Remove trailing spaces from a column.")
+	rootCmd.Flags().BoolVarP(&args.MultiSubnetFailover, "multi-subnet-failover", "M", false, "Provided for backward compatibility. Sqlcmd always optimizes detection of the active replica of a SQL Failover Cluster.")
+
+	rootCmd.Flags().StringVarP(&args.Password, "password", "P", "", "Obsolete. The initial passwords must be set using the SQLCMDPASSWORD environment variable or entered at the password prompt.")
+
+	// Using PersistentFlags() for ErrorSeverityLevel due to data type uint8 , which is not supported in Flags()
+	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "error-severity-level", "V", 0, "Controls the severity level that is used to set the ERRORLEVEL variable on exit.")
+	// ScreenWidth is *int type , The IntVarP() method takes a pointer to an integer variable as its first argument, which it binds to the flag.
+	args.ScreenWidth = new(int)
+	rootCmd.Flags().IntVarP(args.ScreenWidth, "screenwidth", "w", 0, "Specifies the screen width for output. Sets the SQLCMDCOLWIDTH variable.")
```
shueybubbles, line 216: >args.ScreenWidth [](http://example.com/codeflow?start=25&length=16)

is args.ScreenWidth left as nil when the flag is not passed?

cmd/sqlcmd/sqlcmd.go
```diff
@@ -7,61 +7,63 @@ package sqlcmd
...
+	ScreenWidth                 *int
+	TrimSpaces                  bool
+	MultiSubnetFailover         bool
+	Password                    string
 	// Keep Help at the end of the list
-	Help bool `short:"?" help:"-? shows this syntax summary, --help shows modern sqlcmd sub-command help"`
+	Help bool
 }
 
 // Validate accounts for settings not described by Kong attributes
```
shueybubbles, line 69: >accounts [](http://example.com/codeflow?start=12&length=8)

arguments

cmd/sqlcmd/sqlcmd.go
shueybubbles: here's the test for this:
```
set SQLCMDCOLWIDTH=20
16:36:41.50 e:\git>sqlcmd 
1> :listvar
SQLCMDCOLSEP = " "
SQLCMDCOLWIDTH = "20"
```

Notice we preserve the environment variable if the command line flag is not used.

By contrast:
```
16:38:16.97 e:\git>sqlcmd -w 50
1> :listvar
SQLCMDCOLSEP = " "
SQLCMDCOLWIDTH = "50"
```

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,175 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
 
 func Execute(version string) {
-	ctx := kong.Parse(&args, kong.NoDefaultHelp())
-	if args.Version {
-		ctx.Printf("%v\n", version)
-		fmt.Println(localizer.Sprintf("Legal docs and information: aka.ms/SqlcmdLegal"))
-		fmt.Println(localizer.Sprintf("Third party notices: aka.ms/SqlcmdNotices"))
-		os.Exit(0)
+	// Cobra Tryout #2 : This has version and help flags. But they are also getting stucked inside run function without error
+	rootCmd := &cobra.Command{
```
shueybubbles, line 124: this is no longer a "tryout" so remove or edit this comment

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,175 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
+		Use:   "testCommand",
+		Short: "A brief description of your command",
```
shueybubbles, line 126: can we just delete the Use and Short values?

cmd/sqlcmd/sqlcmd_test.go
```diff
@@ -116,19 +115,33 @@ func TestInvalidCommandLine(t *testing.T) {
...
-		{[]string{"-a", "100"}, "test: '-a 100': Packet size has to be a number between 512 and 32767."},
-		{[]string{"-F", "what"}, "--format must be one of \"horiz\",\"horizontal\",\"vert\",\"vertical\" but got \"what\""},
-		{[]string{"-r", "5"}, `--errors-to-stderr must be one of "-1","0","1" but got '\x05'`},
-		{[]string{"-h-4"}, "test: '-h -4': header value must be either -1 or a value between 1 and 2147483647"},
-		{[]string{"-w", "6"}, "test: '-w 6': value must be greater than 8 and less than 65536."},
+		//{[]string{"-E", "-U", "someuser"}, "--use-trusted-connection and --user-name can't be used together"},
+		{[]string{"-a", "100"}, "'-a 100': Packet size has to be a number between 512 and 32767."},
+		//{[]string{"-F", "what"}, "--format must be one of \"horiz\",\"horizontal\",\"vert\",\"vertical\" but got \"what\""},
+		//{[]string{"-r", "5"}, `--errors-to-stderr must be one of "-1","0","1" but got '\x05'`},
+		{[]string{"-h-4"}, "'-h -4': header value must be either -1 or a value between 1 and 2147483647"},
```
shueybubbles, line 122: uncomment all the test entries. All the tests should pass before merging.

pkg/sqlcmd/commands_test.go
```diff
@@ -207,7 +207,7 @@ func TestConnectCommand(t *testing.T) {
 			return []byte{}, nil
 		},
 	}
-	err := connectCommand(s, []string{"someserver -U someuser"}, 1)
+	err := connectCommand(s, []string{"someserver", "-U", "someuser"}, 1)
```
shueybubbles, line 210: >[]string{"someserver", "-U", "someuser"}, 1) [](http://example.com/codeflow?start=26&length=44)

this isn't how the regex for :command works.

pkg/sqlcmd/commands_test.go
```diff
@@ -220,6 +220,7 @@ func TestConnectCommand(t *testing.T) {
 	password := ""
 	username := ""
 	if canTestAzureAuth() {
+		fmt.Printf("Inside canTestAzureAuth")
```
shueybubbles, line 223: >fmt.Printf("Inside canTestAzureAuth") [](http://example.com/codeflow?start=2&length=37)

remove

cmd/sqlcmd/sqlcmd.go
```diff
@@ -118,23 +120,175 @@ func (a SQLCmdArguments) authenticationMethod(hasPassword bool) string {
...
-	// so far sqlcmd prints all the errors itself so ignore it
-	exitCode, _ := run(vars, &args)
-	os.Exit(exitCode)
+	rootCmd.Flags().StringVarP(&args.Password, "password", "P", "", "Obsolete. The initial passwords must be set using the SQLCMDPASSWORD environment variable or entered at the password prompt.")
+
+	// Using PersistentFlags() for ErrorSeverityLevel due to data type uint8 , which is not supported in Flags()
+	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "error-severity-level", "V", 0, "Controls the severity level that is used to set the ERRORLEVEL variable on exit.")
+	screenWidth := rootCmd.Flags().Int("screenwidth", 0, "Specifies the screen width for output")
+	rootCmd.Flags().IntVarP(screenWidth, "w", "w", 0, "Specifies the screen width for output")
+	rootCmd.PersistentPreRun = func(cmd *cobra.Command, argss []string) {
```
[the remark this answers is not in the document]
JyotikaGargg: Is there any specific reason for Lookup.Changed() . I used Lookup while trying but then used the above implementation.
In above implementation , it is fulfilling the condition  :
ScreenWidth field will be correctly set to nil when the value is not passed or set to 0 else will be set to the value passed.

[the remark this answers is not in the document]
shueybubbles: Adding a PersistentPreRun callback creates extra layer of code that seems unnecessary.

cmd/sqlcmd/sqlcmd_test.go
JyotikaGargg: Uncommented all except mutual exclusive flag one.

pkg/sqlcmd/commands.go
```diff
@@ -387,41 +387,29 @@ func listCommand(s *Sqlcmd, args []string, line uint) (err error) {
...
+	database := flags.String("D", "", "database name")
+	username := flags.String("U", "", "user name")
+	password := flags.String("P", "", "password")
+	loginTimeout := flags.String("l", "", "login timeout")
+	authenticationMethod := flags.String("G", "", "authentication method")
 
-	// Fields removes extra whitespace.
-	// Note :connect doesn't support passwords with spaces
-	if _, err = parser.Parse(strings.Fields(cmdLine)); err != nil {
+	err := flags.Parse(args[1:])
```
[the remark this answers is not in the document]
JyotikaGargg: Ahh!! Got this and fixed :)

shueybubbles (approved):
:shipit:

PR #241, opened 2023-02-22
Fix MSI path order

stuartpa:
To avoid issues when ODBC based sqlcmd is already installed on machine, add the go-sqlcmd to the beginning of the user path (also removed the system requirement, to hopefully avoid the UAC prompt)


apoorvdeshmukh (approved):
:shipit:

PR #261, opened 2023-02-28
Establish conn to curret-context if no args given

apoorvdeshmukh:
sqlcmd when invoked with no arguments, executes in backward compat mode and ignore current-context info for establishing connection. This commit changes the default behavior to establish connection to the endpoint associated with current-context if valid user details exits in sqlcmd config file.
This ensures that users can connect to sql server instance created by sqlcmd if no arguments are specified while invoking sqlcmd.

```
\git\go-sqlcmd>.\sqlcmd.exe
1> select @@servername
2> go

--------------------------------------------------------------------------------------------------------------------------------
LAPTOP-VPBI3ASB

(1 row affected)
1> exit

git\go-sqlcmd>.\sqlcmd.exe create mssql --accept-eula
Downloading mcr.microsoft.com/mssql/server:latest
Starting mcr.microsoft.com/mssql/server:latest
Created context "mssql" in "C:\Users\apdeshmukh\.sqlcmd\sqlconfig", configuring user account...
Disabled "sa" account (and rotated "sa" password). Creating user "apdeshmukh"
Now ready for client connections on port 1435

HINT:
  1. Open in Azure Data Studio: sqlcmd open ads
  2. Run a query:               sqlcmd query "SELECT @@version"
  3. Start interactive session: sqlcmd query
  4. View sqlcmd configuration: sqlcmd config view
  5. See connection strings:    sqlcmd config connection-strings
  6. Remove:                    sqlcmd delete


\git\go-sqlcmd>.\sqlcmd.exe
1> select @@servername
2> go

--------------------------------------------------------------------------------------------------------------------------------
4ee585f27077

(1 row affected)
1>
```

### DIFF
--- a/cmd/modern/main.go
+++ b/cmd/modern/main.go
@@ -60,6 +60,7 @@ func initializeEnvVars() {
 		os.Setenv("SQLCMDSERVER", server)
 		os.Setenv("SQLCMDUSER", username)
 		os.Setenv("SQLCMDPASSWORD", password)
+		outputter.Infof("%vUsing current-context: %s", sqlcmd.SqlcmdEol, config.CurrentContextName())
 	}
 }
 
--- a/cmd/modern/main.go
+++ b/cmd/modern/main.go
@@ -41,11 +41,11 @@ func main() {
 			ErrorHandler:   checkErr,
 			HintHandler:    displayHints})}
 	rootCmd = cmdparser.New[*Root](dependencies)
-	initializeEnvVars()
 	if isFirstArgModernCliSubCommand() {
 		cmdparser.Initialize(initializeCallback)
 		rootCmd.Execute()
 	} else {
+		initializeEnvVars()
 		legacyCmd.Execute(version)
 	}
 }
@@ -55,7 +55,7 @@ func main() {
 // if user provides this info via backward compatibility syntax
 func initializeEnvVars() {
 	initializeCallback()
-	if config.CurrentContextName() != "" && config.IsCurrentContextValid() {
+	if config.CurrentContextName() != "" {
 		server, username, password := config.GetCurrentContextInfo()
 		os.Setenv("SQLCMDSERVER", server)
 		os.Setenv("SQLCMDUSER", username)
--- a/cmd/modern/main.go
+++ b/cmd/modern/main.go
@@ -51,16 +51,23 @@ func main() {
 }
 
 // initializeEnvVars intializes SQLCMDSERVER, SQLCMDUSER and SQLCMDPASSWORD
-// if the currentContext is set. This env variables do not have any effect
-// if user provides this info via backward compatibility syntax
+// if the currentContext is set and if these env vars are not already set.
+// In terms of precedence, command line switches/flags take higher precedence
+// than env variables and env variables take higher precedence over config
+// file info.
 func initializeEnvVars() {
 	initializeCallback()
 	if config.CurrentContextName() != "" {
 		server, username, password := config.GetCurrentContextInfo()
-		os.Setenv("SQLCMDSERVER", server)
-		os.Setenv("SQLCMDUSER", username)
-		os.Setenv("SQLCMDPASSWORD", password)
-		outputter.Infof("%vUsing current-context: %s", sqlcmd.SqlcmdEol, config.CurrentContextName())
+		if os.Getenv("SQLCMDSERVER") == "" {
+			os.Setenv("SQLCMDSERVER", server)
+		}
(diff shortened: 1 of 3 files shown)


stuartpa:
> 		legacyCmd.Execute(version)

Move this call to the else statement below (right now we only need this envvars when running in legacy mode)

---
Refers to: cmd/modern/main.go:49 in 2f1a322. [](commit_id = 2f1a322c68cd1c6f7c0a6910b4635bca4134ad7b, deletion_comment = False)

internal/config/context.go
```diff
@@ -94,6 +94,47 @@ func CurrentContext() (endpoint Endpoint, user *User) {
...
+	if user.AuthenticationType == "basic" {
+		password = decryptCallback(
+			user.BasicAuth.Password,
+			user.BasicAuth.PasswordEncrypted,
+		)
+	}
+	return
+}
+
+// isCurrentContextValid returns true if it has valid information associated
```
stuartpa, line 112: I (captial)

stuartpa (changes requested):
Changes as discussed earlier

internal/config/context.go
apoorvdeshmukh: Removed this function as it was not needed.

apoorvdeshmukh:
> Changes as discussed earlier

Incorporated the changes which we discussed earlier. 
Now we are setting only those env variables for which info is available in context such as endpoint and skipping others such as user details if not available (in case context is to be used for trusted authentication)

apoorvdeshmukh:
> > ```
> >   legacyCmd.Execute(version)
> > ```
> 
> Move this call to the else statement below (right now we only need this envvars when running in legacy mode)
> 
> Refers to: cmd/modern/main.go:49 in [2f1a322](https://github.com/microsoft/go-sqlcmd/commit/2f1a322c68cd1c6f7c0a6910b4635bca4134ad7b). [](commit_id = [2f1a322](https://github.com/microsoft/go-sqlcmd/commit/2f1a322c68cd1c6f7c0a6910b4635bca4134ad7b), deletion_comment = False)

Fixed

cmd/modern/main.go
```diff
@@ -45,10 +45,25 @@ func main() {
...
 }
 
+// initializeEnvVars intializes SQLCMDSERVER, SQLCMDUSER and SQLCMDPASSWORD
+// if the currentContext is set. This env variables do not have any effect
+// if user provides this info via backward compatibility syntax
+func initializeEnvVars() {
+	initializeCallback()
+	if config.CurrentContextName() != "" {
+		server, username, password := config.GetCurrentContextInfo()
+		os.Setenv("SQLCMDSERVER", server)
```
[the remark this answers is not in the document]
apoorvdeshmukh: Good point.  In case user already has env variables set and also has current-context set, then shouldn't the current-context info take precedence here?  If current-context doesn't have user info, then I agree that existing env variables (SQLCMDUSER, SQLCMDPASSWORD) should be left untouched.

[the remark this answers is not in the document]
stuartpa: I don't feel great about this new behavior taking precedence, for existing scripts etc.  If a user has set this env vars, I think we should honor that.  And only set them here if they are not already set.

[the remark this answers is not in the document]
shueybubbles: environment variables should always take precedence over config files, and command line switches take precedence over environment variables.

internal/config/context.go
```diff
@@ -94,6 +96,23 @@ func CurrentContext() (endpoint Endpoint, user *User) {
 	return
 }
 
+// GetCurrentContextInfo returns endpoint and basic auth info
+// associated with current context
+func GetCurrentContextInfo() (server string, username string, password string) {
+	endpoint, user := CurrentContext()
+	server = fmt.Sprintf("%s,%d", endpoint.Address, endpoint.Port)
```
shueybubbles, line 103: >,%d [](http://example.com/codeflow?start=25&length=3)

do endpoints always have a port? What about people using non-tcp connections?

shueybubbles: how about people using named instances?

apoorvdeshmukh: As per the help summary for add-endpoint, it needs --address and --port defaults to 1433 if, not specified

shueybubbles: pls open an issue to track support for non-tcp connections, as they haven't gone away from real enterprise deployments yet

stuartpa (approved):
:shipit:

PR #269, opened 2023-03-01
Fix Package pipeline

stuartpa:
Move to EsrpCodeSigning@2


shueybubbles:
>           "OperationCode" : "LinuxSign",

i didn't know we had a "LinuxSign"
Is there a version that works on the executable we should be using?

---
Refers to: build/azure-pipelines/package-common-create.yml:125 in 6d5906b. [](commit_id = 6d5906bd96ddf224d900483f5190557a2d9830ee, deletion_comment = False)

shueybubbles (approved):
:shipit:

PR #270, opened 2023-03-01
Choco Package for v0.15.0

stuartpa:


release/windows/choco/sqlcmd.nuspec
```diff
@@ -3,11 +3,10 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>sqlcmd</id>
-    <version>0.8.1</version>
+    <version>0.15.0</version>
```
shueybubbles, line 6: >>0.15.0< [](http://example.com/codeflow?start=12&length=8)

can we change the pipeline that packs the package get the version from the current tag instead of hard coding it here? Then line 6 and line 20 can reference a variable like $version$ and the pipeline will pass version as a parameter to the pack.

shueybubbles (approved):
:shipit:

shueybubbles:
looks like you need to merge

shueybubbles (approved):

PR #271, opened 2023-03-01
create a sqlcmd style linter

### LINKED ISSUE #214: Enforce use of assert.Equalf(t, ...) pattern instead of if != then { t.Errorf(...) } pattern
Code generators (e.g. goland) by default seem to be generating test code templates using the t.Errorf pattern, which is verbose (multiple lines), when the assert.Equalf(t, ...) pattern is more concise.

How do we prevent the t.Errorf pattern from being merged?  What do we need to put in our pipelines to cause a pipeline break if patterns (especially default ones like t.Errorf) are not the patterns we want merged as future PRs.

shueybubbles:
Fixes #214 
- Iteration 1 adds a basic linter skeleton with one linter that flags use of non-`assert` test failure methods like `t.Fatalf`
- Adds use of the linter to `build\build.cmd` to print out any warnings. Once we are clean we can convert the linter run to a step in the PR validation pipeline.
- Iteration 2 will add enforcement of package import/hierarchy rules.

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -249,6 +249,38 @@ SOFTWARE.
 
 ```
 
+## github.com/cespare/xxhash/v2
+
+* Name: github.com/cespare/xxhash/v2
+* Version: v2.1.1
+* License: [MIT](https://github.com/cespare/xxhash/blob/v2.1.1/LICENSE.txt)
+
+```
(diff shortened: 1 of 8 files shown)


stuartpa (approved):
:shipit:

PR #278, opened 2023-03-03
Add console syntax coloring

shueybubbles:
- use `SQLCMDCOLORSCHEME` variable to identify the style/scheme name that matches one defined by the [chroma styles]<https://github.com/alecthomas/chroma/styles> package
- Add a `:list color` command to print samples of TSQL syntax in each scheme
- Colorizes query results and errors as well

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -159,6 +159,35 @@ SOFTWARE.
 
 ```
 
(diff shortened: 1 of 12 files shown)


stuartpa (approved):
:shipit:

PR #280, opened 2023-03-03
Add --database to sqlcmd config cs

### LINKED ISSUE #277: Add --database option to the connection-string command
Right now the connection strings are always generated for the `master` database. Would be great to have an option `--database` or `-d` to have the connection strings created with the specified database name

stuartpa:
Closes #277

### DIFF
--- a/cmd/modern/root/config/connection-strings.go
+++ b/cmd/modern/root/config/connection-strings.go
@@ -17,6 +17,8 @@ import (
 // ConnectionStrings implements the `sqlcmd config connection-strings` command
 type ConnectionStrings struct {
 	cmdparser.Cmd
+
+	database string
 }
 
 func (c *ConnectionStrings) DefineCommand(...cmdparser.CommandOptions) {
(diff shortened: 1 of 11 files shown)


cmd/modern/root/config/connection-strings.go
```diff
@@ -36,6 +38,13 @@ func (c *ConnectionStrings) DefineCommand(...cmdparser.CommandOptions) {
 	}
 
 	c.Cmd.DefineCommand(options)
+
+	c.AddFlag(cmdparser.FlagOptions{
+		String:        &c.database,
+		Name:          "database",
+		DefaultString: "master",
```
shueybubbles, line 45: wouldn't the default be empty so the user gets their default database?

shueybubbles: I think I need some clarification of the intent of this function. 
Pls let me know if my understanding is correct:

If the user creates a context in the config file, it could be in one of 2 states:
- no database set
- a specific database is set

When the user runs the ConnectionStrings command, if they don't pass -d, we should provide whatever database is in the context, which could mean not providing any specific database name. If they do pass -d, we would override the database name from the context.

yorek: > if they don't pass -d, we should provide whatever database is in the context, which could mean not providing any specific database name. 

This would be the best option IMHO. By default, you get the database the has been set in the context. `-d` can override that

stuartpa: Yeah, agree.  We don't actually set the database in the context, we set in on the login as DEFAULT_DATABASE.  So this would mean we need sqlcmd to connect to the container/sqlservr, get the DB_NAME() and use this by default (when -d is not passed in)

JyotikaGargg (approved):

PR #286, opened 2023-03-07
Allow sqlcmd delete to succeed if container has already been deleted

### LINKED ISSUE #282: `sqlcmd delete` fails for a context that has a container, and the container has been manually deleted already
We could do a couple of things

1. Allow the delete to succeed, and log a warning
2. Fail the delete, and require a --force  (but right now --force is bound to the scenario around not deleteing a context if user data is present in the contianer)

stuartpa:
This PR allows the `sqlcmd delete` to succeed even if the user has manually deleted the container using something like `docker rm`.  sqlcmd will log a warning if it finds the container has already been removed, and will continue to delete the context (and endpoint/user) and will complete successfullly

Closes #282

### DIFF
--- a/cmd/modern/root/uninstall.go
+++ b/cmd/modern/root/uninstall.go
@@ -124,7 +124,7 @@ func (c *Uninstall) run() {
 				err = controller.ContainerRemove(id)
 				c.CheckErr(err)
 			} else {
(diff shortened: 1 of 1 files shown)


shueybubbles (approved):
:shipit:

PR #290, opened 2023-03-08
Add support for customized database name with --using

### LINKED ISSUE #216: Support customized T/SQL database name with --using flag
It would be nice for people to be able to set the T/SQL name of their database when using the --using flag, at the moment the name is set to the .bak file name, e.g. AdventureWorksLT.  The suggestion is to use the following "," format, https://aka.ms/AdventureWorksLT.bak,adventure_works, where the name of the database comes after the comma, e.g. adventure_works in this case.

apoorvdeshmukh:
This commit adds support for specifying database name with `--using` option by specifying `,DbName` at the end of the URL
This PR closes #216 

Example without specifying customized database name, in which case Dbname defaults to the filename without extension

```
\git\go-sqlcmd>sqlcmd create mssql --accept-eula --using https://aka.ms/AdventureWorksLT.bak
Downloading mcr.microsoft.com/mssql/server:latest
Starting mcr.microsoft.com/mssql/server:latest
Created context "mssql" in "C:\Users\apdeshmukh\.sqlcmd\sqlconfig", configuring user account...
Disabled "sa" account (and rotated "sa" password). Creating user "apdeshmukh"Downloading AdventureWorksLT.bak from aka.ms
Restoring database AdventureWorksLT
Processed 840 pages for database 'AdventureWorksLT', file 'AdventureWorksLT2012_Data' on file 1.
Processed 2 pages for database 'AdventureWorksLT', file 'AdventureWorksLT2012_Log' on file 1.
Converting database 'AdventureWorksLT' from version 904 to the current version 957.
Database 'AdventureWorksLT' running the upgrade step from version 904 to version 905.
Database 'AdventureWorksLT' running the upgrade step from version 955 to version 956.
Database 'AdventureWorksLT' running the upgrade step from version 956 to version 957.
RESTORE DATABASE successfully processed 842 pages in 0.038 seconds (173.005 MB/sec).
Now ready for client connections on port 1435HINT:
  1. Open in Azure Data Studio: sqlcmd open ads
  2. Run a query:               sqlcmd query "SELECT @@version"  3. Start interactive session: sqlcmd query
  4. View sqlcmd configuration: sqlcmd config view
  5. See connection strings:    sqlcmd config connection-strings
  6. Remove:                    sqlcmd delete
\git\go-sqlcmd>.\sqlcmd.exe query
1> SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE='BASE TABLE'
2> go
TABLE_CATALOG                                                                                                                    TABLE_SCHEMA                                                                                                                     TABLE_NAME
                                                                                                           TABLE_TYPE
-------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- ----------AdventureWorksLT                                                                                                                 dbo                                                                                                                              ErrorLog
                                                                                                           BASE TABLE
AdventureWorksLT                                                                                                                 dbo                                                                                                                              BuildVersion
                                                                                                           BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          Address
                                                                                                           BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          Customer
                                                                                                           BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          CustomerAddress                                                                                                                  BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          Product
                                                                                                           BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          ProductCategory                                                                                                                  BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          ProductDescription                                                                                                               BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          ProductModel
                                                                                                           BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          ProductModelProductDescription                                                                                                   BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          SalesOrderDetail                                                                                                                 BASE TABLE
AdventureWorksLT                                                                                                                 SalesLT                                                                                                                          SalesOrderHeader                                                                                                                 BASE TABLE
(12 rows affected)
```
Example with customized database name, in which case Dbname is set to user provided value

```
\git\go-sqlcmd>sqlcmd create mssql --accept-eula --using https://aka.ms/AdventureWorksLT.bak,WorldWideImporters
Downloading mcr.microsoft.com/mssql/server:latest
Starting mcr.microsoft.com/mssql/server:latest
Created context "mssql" in "C:\Users\apdeshmukh\.sqlcmd\sqlconfig", configuring user account...
Disabled "sa" account (and rotated "sa" password). Creating user "apdeshmukh"Downloading AdventureWorksLT.bak from aka.ms
Restoring database WorldWideImporters
Processed 840 pages for database 'WorldWideImporters', file 'AdventureWorksLT2012_Data' on file 1.
Processed 2 pages for database 'WorldWideImporters', file 'AdventureWorksLT2012_Log' on file 1.
Converting database 'WorldWideImporters' from version 904 to the current version 957.
Database 'WorldWideImporters' running the upgrade step from version 904 to version 905.
Database 'WorldWideImporters' running the upgrade step from version 954 to version 955.
Database 'WorldWideImporters' running the upgrade step from version 955 to version 956.
Database 'WorldWideImporters' running the upgrade step from version 956 to version 957.
RESTORE DATABASE successfully processed 842 pages in 0.038 seconds (173.005 MB/sec).
Now ready for client connections on port 1435HINT:
  1. Open in Azure Data Studio: sqlcmd open ads
  2. Run a query:               sqlcmd query "SELECT @@version"  3. Start interactive session: sqlcmd query
  4. View sqlcmd configuration: sqlcmd config view
  5. See connection strings:    sqlcmd config connection-strings
  6. Remove:                    sqlcmd delete
\git\go-sqlcmd>.\sqlcmd.exe query
1> select db_name()
2> go
--------------------------------------------------------------------------------------------------------------------------------WorldWideImporters
(1 row affected)
1> SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE='BASE TABLE'
2> go
TABLE_CATALOG                                                                                                                    TABLE_SCHEMA                                                                                                                     TABLE_NAME                                                                                                                       TABLE_TYPE
-------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- -------------------------------------------------------------------------------------------------------------------------------- ----------WorldWideImporters                                                                                                               dbo                                                                                                                              ErrorLog                                                                                                                         BASE TABLE
WorldWideImporters                                                                                                               dbo                                                                                                                              BuildVersion                                                                                                                     BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          Address                                                                                                                          BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          Customer                                                                                                                         BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          CustomerAddress                                                                                                                  BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          Product                                                                                                                          BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          ProductCategory                                                                                                                  BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          ProductDescription                                                                                                               BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          ProductModel                                                                                                                     BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          ProductModelProductDescription                                                                                                   BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          SalesOrderDetail                                                                                                                 BASE TABLE
WorldWideImporters                                                                                                               SalesLT                                                                                                                          SalesOrderHeader                                                                                                                 BASE TABLE
(12 rows affected)

```


cmd/modern/root/install/mssql-base.go
```diff
@@ -444,31 +450,48 @@ CHECK_POLICY=OFF`
 	}
 }
 
+func getDbNAmeIfExists(usingDbUrl string) string {
```
stuartpa, line 453: typo

apoorvdeshmukh: Fixed it

shueybubbles:
please add tests. There's a pretty big matrix of input to cover here and the tests need to show what's valid and what's not. We need to handle edge cases that could lead to sql injection if we don't escape the name, etc.
Can I pass a local file path like --using "c:\path with a space and a comm,a\path","db space and co,mma"  ?

shueybubbles (changes requested):
:clock1:

shueybubbles:
do we want to allow the user to provide the db name as a bracketed identifier? 
eg 
`--using [https://someurl/somefile.bak,[my](https://someurl/somefile.bak,[my) database]`


---
In reply to: [1460392606](https://github.com/microsoft/go-sqlcmd/pull/290#issuecomment-1460392606) [](http://example.com/codeflow?ancestors=1460392606)

shueybubbles:
if there's a space in the name or the path will the user have to surround the entire parameter with quotes?

---
In reply to: [1460405555](https://github.com/microsoft/go-sqlcmd/pull/290#issuecomment-1460405555) [](http://example.com/codeflow?ancestors=1460405555,1460392606)

cmd/modern/root/install/mssql-base.go
```diff
@@ -444,31 +450,48 @@ CHECK_POLICY=OFF`
...
 func (c *MssqlBase) downloadAndRestoreDb(
 	controller *container.Controller,
 	containerId string,
 	userName string,
 ) {
 	output := c.Cmd.Output()
+	databaseName := getDbNameIfExists(c.usingDatabaseUrl)
+	databaseUrl := c.usingDatabaseUrl
+	if databaseName != "" {
+		databaseUrl = strings.Split(c.usingDatabaseUrl, ",")[0]
```
stuartpa, line 472: Don't use Split, just take everything after the first comma

stuartpa:
Yes, add an example with --using with a space in the db name (it works, we tested it)

---
In reply to: [1460405555](https://github.com/microsoft/go-sqlcmd/pull/290#issuecomment-1460405555) [](http://example.com/codeflow?ancestors=1460405555,1460392606)

cmd/modern/root/install/mssql-base.go
shueybubbles: file names, folder names, and database names can have commas in them (yes SSMS gets real support cases because of this). Make sure that cobra is passing correctly delimited strings and that this code is aware of whether the comma is part of the path or is the real separator between path and name

shueybubbles:
> 		String:        &c.usingDatabaseUrl,

Is there a way we could change this flag type to a `StringSlice` and restrict it to only being used once? That could let cobra do the heavy lifting of delimiting the components by the comma, and our code could potentially allow more parameters in the future. 

I think the default behavior of StringSlice is to allow the same flag to be used multiple times like

```
-u f1,f2 -u f3
```



---
Refers to: cmd/modern/root/install/mssql-base.go:192 in 246bd0e. [](commit_id = 246bd0ee6d8d80eddd658fa719dd6c91d26f6859, deletion_comment = False)

apoorvdeshmukh:
> please add tests. There's a pretty big matrix of input to cover here and the tests need to show what's valid and what's not. We need to handle edge cases that could lead to sql injection if we don't escape the name, etc. Can I pass a local file path like --using "c:\path with a space and a comm,a\path","db space and co,mma" ?

We do not support filesystem paths at the moment. Only http and https. But I will add tests to check what happens if such input is intentionally passed.

apoorvdeshmukh:
> do we want to allow the user to provide the db name as a bracketed identifier? eg
> 
> ```
> --using https://someurl/somefile.bak,[my database]
> ```
> 
> In reply to: [1460392606](https://github.com/microsoft/go-sqlcmd/pull/290#issuecomment-1460392606)

I checked with @stuartpa and for now we will only support `URL,DbName` where `,DbName` is optional. 
If there are spaces or commas in the DbName then we would expect user to surround the entire argument of --using with `""`
E.g. `--using "https://aka.ms/AdventureWorksLT.bak,My Db"`

shueybubbles:
> SET @sql = 'RESTORE DATABASE [%s] FROM DISK = ''%s/%s'' WITH '

if databaseName has a ] in it this query becomes a vector for sql injection.
Similarly if any of the other parameters have a ' in their value this query will go awry.

Either use a properly parameterized query (go-mssqldb supports parameterized queries) or escape everything appropriately.

---
Refers to: cmd/modern/root/install/mssql-base.go:530 in 7a37b69. [](commit_id = 7a37b697417c615cee774b5bc6d61b9e0a2ea236, deletion_comment = False)

stuartpa:
We want to incrementally improve the quality of the experience here.  There is a quality issue with the experience right now, in that folks don't want to be stuck with the name of the .bak file as the database name.  It would improve the experience if we allow a database name to be specified.  In the limit we should support a database name that covers the entire input space for T/SQL identifiers, but we don't have to do that in a single PR.  It is fine in this PR to limit the input space for the database name so SQL injection is not possible, and enter an issue to widen the input space.  What we don't want to see is we end up reducing quality by not thinking through the complete solution for all T/SQL identifier input space in one PR.

cmd/modern/root/install/mssql.go
```diff
@@ -29,6 +29,9 @@ func (c *Mssql) DefineCommand(...cmdparser.CommandOptions) {
 			{
 				Description: "Create SQL Server, download and attach AdventureWorks sample database",
 				Steps:       []string{"sqlcmd create mssql --using https://aka.ms/AdventureWorksLT.bak"}},
+			{
+				Description: "Create SQL Server, download and attach AdventureWorks sample database with different database name",
+				Steps:       []string{"sqlcmd create mssql --using https://aka.ms/AdventureWorksLT.bak,MyDatabaseName"}},
```
stuartpa, line 34: Let's make the example:

--using [https://aka.ms/AdventureWorksLT.bak,adventureworks](https://aka.ms/AdventureWorksLT.bak,adventureworks)

All the system databases are lower case, to customers tend to lean towards lower case (snake case for spacing).

cmd/modern/root/install/mssql-base_test.go
```diff
@@ -0,0 +1,32 @@
...
+		input          string
+		expectedOutput string
+	}
+
+	tests := []test{
+		// Positive Testcases
+		{"https://example.com,myDbName", "myDbName"},
+		{"https://example.com,myDb Name", "myDb Name"},
+		{"https://example.com,myDb Na,me", "myDb Na,me"},
+		{"https://example.com,[myDb Na,me]", "[myDb Na,me]]"},
```
shueybubbles, line 21: >"[myDb Na,me]]"}, [](http://example.com/codeflow?start=39&length=17)

if this function is returning the string that's used for inserting into the TSQL, rename the function `getDbNameForScript` to make it clear that's not the actual name.

cmd/modern/root/install/mssql-base_test.go
```diff
@@ -0,0 +1,32 @@
...
+		{"https://example.com,[myDb Na'me]", "[myDb Na''me]]"},
```
shueybubbles, line 22: >'' [](http://example.com/codeflow?start=48&length=2)

The '``should only be doubled if this string is also being injected into the EXEC  dynamic SQL parameter.

shueybubbles:
We should be able to write a functional test for this in the ADO pipeline. ADO hosted agents have docker on them, so the pipeline should be able to use the sqlcmd it builds to run some permutations on `sqlcmd create --using` and verify the E2E scenario.
The unit test is insufficient to prove we are doing the right thing with the user input all the way through the download and rename process.

---
In reply to: [1462251401](https://github.com/microsoft/go-sqlcmd/pull/290#issuecomment-1462251401) [](http://example.com/codeflow?ancestors=1462251401,1462251401)

cmd/modern/root/install/mssql-base.go
```diff
@@ -462,31 +468,51 @@ CHECK_POLICY=OFF`
...
 
-	u, err := url.Parse(c.usingDatabaseUrl)
+	u, err := url.Parse(databaseUrl)
 	c.CheckErr(err)
-	_, file := filepath.Split(c.usingDatabaseUrl)
+	_, file := filepath.Split(databaseUrl)
 	fileNameWithNoExt := strings.TrimSuffix(file, filepath.Ext(file))
-
+	if databaseName == "" {
+		databaseName = fileNameWithNoExt
```
shueybubbles, line 501: >fileNameWithNoEx [](http://example.com/codeflow?start=17&length=16)

can the file name portion of a URL also have ] characters in it?

cmd/modern/root/install/mssql-base.go
```diff
@@ -462,31 +468,51 @@ CHECK_POLICY=OFF`
...
 	controller.DownloadFile(
 		containerId,
-		c.usingDatabaseUrl,
+		databaseUrl,
 		temporaryFolder,
 	)
 
 	// Restore database from file
-	output.Infof("Restoring database %s", fileNameWithNoExt)
+	output.Infof("Restoring database %s", databaseName)
```
shueybubbles, line 515: >databaseName [](http://example.com/codeflow?start=39&length=12)

any name we display to the user should not be escaped

cmd/modern/root/install/mssql-base.go
apoorvdeshmukh: Yes. It is possible. (At least on windows platform)

apoorvdeshmukh: Have escaped DbName generated from file name portion as well

cmd/modern/root/install/mssql-base_test.go
apoorvdeshmukh: This is indeed the case in our code.

apoorvdeshmukh:
> We should be able to write a functional test for this in the ADO pipeline. ADO hosted agents have docker on them, so the pipeline should be able to use the sqlcmd it builds to run some permutations on `sqlcmd create --using` and verify the E2E scenario. The unit test is insufficient to prove we are doing the right thing with the user input all the way through the download and rename process.
> 
> In reply to: [1462251401](https://github.com/microsoft/go-sqlcmd/pull/290#issuecomment-1462251401)

Created #306 to add functional tests for E2E testing scenarios

cmd/modern/root/install/mssql-base.go
apoorvdeshmukh: > > fileNameWithNoEx
> 
> can the file name portion of a URL also have ] characters in it?

I was wrong about this. The RFC for URL states that a filename can only contain `-`,`.`,`_`,`~`
So a filename in URL can never have `]`. The filename, if it contains the `]` character then it's most likely to be in its encoded form.
[RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-2.3)

shueybubbles: sure, which is why we should be decoding the Url properly not just scanning it for a comma

cmd/modern/root/install/mssql-base.go
```diff
@@ -408,7 +410,14 @@ func (c *MssqlBase) validateUsingUrlExists() {
...
 	// support .bacpacs and .mdfs in the future
-	if _, file := filepath.Split(c.usingDatabaseUrl); filepath.Ext(file) != ".bak" {
+	databaseName := parseDbName(c.usingDatabaseUrl)
+	databaseUrl := c.usingDatabaseUrl
+	if databaseName != "" {
+		argLen := len(c.usingDatabaseUrl)
+		dbNameLen := len(databaseName)
+		databaseUrl = c.usingDatabaseUrl[0:(argLen - dbNameLen - 1)]
+	}
+	if _, file := filepath.Split(databaseUrl); filepath.Ext(file) != ".bak" {
```
shueybubbles, line 420: >filepath.Split(databaseUrl);  [](http://example.com/codeflow?start=15&length=29)

shouldn't we be using the last segment of url.Path as the value of file? The file name in the URL is going to be url encoded.

cmd/modern/root/install/mssql-base.go
```diff
@@ -468,32 +477,55 @@ CHECK_POLICY=OFF`
...
+	return ""
+}
+
 func (c *MssqlBase) downloadAndRestoreDb(
 	controller *container.Controller,
 	containerId string,
 	userName string,
 ) {
 	output := c.Cmd.Output()
+	databaseName := parseDbName(c.usingDatabaseUrl)
```
shueybubbles, line 504: >parseDbName(c.usingDat [](http://example.com/codeflow?start=17&length=22)

it feels like there's an opportunity to consolidate some of this code with the validation code by changing parseDbName to return the database name for both cases - when there's a `,dbname` and when there isn't.
Then the unit test for parseDbName can validate the full range of  URLs with and without escaped characters in the file names in the url etc.

cmd/modern/root/install/mssql-base_test.go
```diff
@@ -0,0 +1,34 @@
+package install
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetDbNameIfExists(t *testing.T) {
```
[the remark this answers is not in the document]
apoorvdeshmukh: Yes added a test case for this scenario.

cmd/modern/root/install/mssql-base.go
```diff
@@ -468,32 +477,55 @@ CHECK_POLICY=OFF`
...
+	databaseUrl := c.usingDatabaseUrl
+	if databaseName != "" {
+		databaseUrl = strings.Split(c.usingDatabaseUrl, ",")[0]
+	}
 
-	u, err := url.Parse(c.usingDatabaseUrl)
+	u, err := url.Parse(databaseUrl)
 	c.CheckErr(err)
-	_, file := filepath.Split(c.usingDatabaseUrl)
+	_, file := filepath.Split(databaseUrl)
```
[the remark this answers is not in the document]
apoorvdeshmukh: This is the existing code. u.Path is generally URL decoded. I was not sure if there will be any issues if the decoded file name has spaces and special characters so kept the existing raw argument which is subsequently passed to wget and wget seems to handle encoded strings well.

cmd/modern/root/install/mssql-base.go
apoorvdeshmukh: Consolidated the code to set dbName to filename when dbName is not specified.

cmd/modern/root/install/mssql-base.go
```diff
@@ -468,17 +480,55 @@ CHECK_POLICY=OFF`
...
 	controller *container.Controller,
 	containerId string,
 	userName string,
 ) {
 	output := c.Cmd.Output()
+	databaseName := parseDbName(c.usingDatabaseUrl)
+	databaseUrl := extractUrl(c.usingDatabaseUrl)
 
-	u, err := url.Parse(c.usingDatabaseUrl)
+	u, err := url.Parse(databaseUrl)
```
shueybubbles, line 529: >Parse(databaseUrl) [](http://example.com/codeflow?start=15&length=18)

parsedbName is already calling url.Parse so why do we need to do it again? 
Can parseDbName return all the relevant Url attributes ?

cmd/modern/root/install/mssql-base.go
```diff
@@ -535,12 +585,12 @@ WHERE IsPresent = 1
...
 EXEC(@sql)`
 
-	c.query(fmt.Sprintf(text, temporaryFolder, file, fileNameWithNoExt, temporaryFolder, file))
+	c.query(fmt.Sprintf(text, temporaryFolder, file, databaseName, temporaryFolder, file))
 
 	alterDefaultDb := fmt.Sprintf(
 		"ALTER LOGIN [%s] WITH DEFAULT_DATABASE = [%s]",
 		userName,
-		fileNameWithNoExt)
+		databaseName)
```
shueybubbles, line 593: >databaseName) [](http://example.com/codeflow?start=2&length=13)

the escaping of `'` characters for this query will be incorrect.
This query isn't injecting the name into a string for dynamic sql so only the ] should be escaped.

As I mentioned before - the identifier should be escaped only at the point of insertion, using the appropriate escaping mechanism. 
A functional test would catch this bug.

cmd/modern/root/install/mssql-base.go
apoorvdeshmukh: This was only needed for printing the u.Hostname()
I think we need not return URL attributes just for this case since we are already validating the URL as part of validateUsingUrlExists() which gets called before even we download the file.
I will remove this code which requires us to parse the URL again.

shueybubbles (approved):
:shipit:

PR #293, opened 2023-03-09
Require --password-encryption for add-user

stuartpa:
Require user supply `--password-encryption none | dpapi | etc.` flag when doing `sqlcmd config add-user` (which could be used for a remote endpoint), and optionally supply it for `sqlcmd create` (in this case it is always local, so in the same trust boundary)

This is a breaking change if any user of go-sqlcmd has previously used boolean --encrypt-password flag (which has been removed in this PR), and set it to true (but I am not aware of any usage of this flag so far).  If any user does hit this breaking change, the fix is simple: 
  Change the line in your sqlconfig
    password-encrypted: true
  to
    password-encryption: dpapi

(the error the user will get in ADS (when running `sqlcmd open ads`) if this happens is: The value's length for key 'Password' exceeds its limit of '128'. 

I have manually tested that this change works where a sqlconfig file was created with an old go-sqlcmd, for the normal case, where the user has just run `sqlcmd create mssql`, in this case there is temporary code to cope with the missing "password-encryption: none".

e.g.

sqlcmd config add-user --username sa --password-encryption none

or

sqlcmd config add-user --username sa --password-encryption dpapi
sqlcmd config add-user --username sa --password-encryption keychain
etc.

Change the sqlconfig file from:

users:
- name: stuartpa@mssql
  authentication-type: basic
  basic-auth:
    username: stuartpa
    password-encrypted: false
    password: REDACTED

to:

users:
- name: stuartpa@mssql
  authentication-type: basic
  basic-auth:
    username: stuartpa
    password-encryption: none
    password: REDACTED

or

users:
- name: stuartpa@mssql
  authentication-type: basic
  basic-auth:
    username: stuartpa
    password-encryption: dpapi
    password: REDACTED

etc.


internal/cmdparser/cmd.go
```diff
@@ -54,35 +71,77 @@ func (c *Cmd) AddFlag(options FlagOptions) {
 			panic("Only provide one type")
 		}
 		if options.Shorthand == "" {
-			c.command.PersistentFlags().IntVar(
-				options.Int,
-				options.Name,
-				options.DefaultInt,
-				options.Usage)
+			if options.Persistent {
```
shueybubbles, line 74: >.Persistent [](http://example.com/codeflow?start=13&length=11)

How about creating a wrapper over `c.command.[Persistent]Flags()` that implements the flags interface and handles the if Persistent check so we don't need these if/else statements everywhere

shueybubbles: the calling code could look like this
```
if options.Shorthand == "" {
getFlags(c.Command).IntVar(
```

shueybubbles:
the content of this Pr doesn't match the description

stuartpa:
> the content of this Pr doesn't match the description

I'm working on it! :-) (It's still in "draft" state, but thanks for the feedback so far)

internal/cmdparser/cmd.go
stuartpa: I've moved this code out of this PR, will do this in it's own PR.

apoorvdeshmukh (approved):

cmd/modern/root/config/add-user.go
```diff
@@ -18,35 +19,48 @@ import (
...
-		Run: c.run}
+	}
+
+	if runtime.GOOS == "windows" {
+		examples = append(examples, cmdparser.ExampleOptions{
+			Description: "Add a user using Windows Data Protection API to encrypt password in sqlconfig",
+			Steps: []string{
+				fmt.Sprintf(`%s SQLCMD_PASSWORD=<placeholderpassword>`, pal.CreateEnvVarKeyword()),
+				"sqlcmd config add-user --name my-user --username user1 --password-encryption dpapi",
+				fmt.Sprintf(`%s SQLCMD_PASSWORD=`, pal.CreateEnvVarKeyword()),
```
shueybubbles, line 54: >	fmt.Sprintf(`%s SQLCMD_PASSWORD=`, pal.CreateEnvVarKeyword()), [](http://example.com/codeflow?start=3&length=63)

it'd be useful to add a 
`func setEnvironmentVar(var string, val string) string` to `pal` so you don't have to use fmt.Sprintf everywhere and to handle any potential need to escape the value in a per-platform manner

stuartpa: Added (so we don't lose this feedback in this merged PR):

https://github.com/microsoft/go-sqlcmd/issues/307

PR #319, opened 2023-04-06
Round out --use scenario (add "use" sub-command, add local file, add .bacpac / .mdf & .ldf / compressed file support)

stuartpa:
This round out the "use" scenario. sqlcmd now supports

1. --use now supports local and remote files
```sqlcmd create mssql --use [<local file>>|<<remote file>]```

3. --use now supports .mdfs/.bacpac and .bak files 
```sqlcmd create mssql --use [.mdf | .bacpac | *.bak]```

4. --use now support compressed file formats (that also allows a .mdf/.ldf file pair)
```sqlcmd create mssql --use [.7z | .zip | .tar]  # containing any of the above (inc. a .mdf/.ldf pair)```

5. the new `sqlcmd use` subcommand allows all of the above to an already existing container
```sqlcmd use [<local file>|<remote file>]```

Examples:
```
sqlcmd install mssql --accept-eula

# Restore a backup
sqlcmd use https://aka.ms/AdventureWorksLT.bak

# Restore a Dacfx .bacpac
sqlcmd use https://github.com/Microsoft/sql-server-samples/releases/download/wide-world-importers-v1.0/WideWorldImportersDW-Full.bacpac

# Attach a compressed .mdf/.ldf
sqlcmd use https://downloads.brentozar.com/StackOverflow2010.7z
```

### CODE REVIEW

**Comment by shueybubbles** [comment]
What about supporting non-public URLs (SAS etc)? Does it download the file then discover its type similar to how you'd have to examine a zip to know its content is?

File: cmd/modern/root/use.go [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+
+		if !controller.ContainerRunning(id) {
+			output.FatalfWithHintExamples([][]string{
+				{"Start container for current context", "sqlcmd start"},
+			}, "Container for current context is not running")
+		}
+
+		endpoint, user := config.CurrentContext()
+
+		c.sql = sql.New(sql.SqlOptions{UnitTesting: false})
```
shueybubbles, line 73: >UnitTesting: false} [](http://example.com/codeflow?start=33&length=19)

isn't false the default value?
We shouldn't need to have product code give hints about unit testing. Test hooks should be available for tests to inject things.

cmd/modern/root/use.go
```diff
@@ -0,0 +1,107 @@
...
+		useDatabase.BringOnline(
+			c.sql.Query,
+			user.BasicAuth.Username,
+			secret.Decode(user.BasicAuth.Password, user.BasicAuth.PasswordEncryption),
+		)
+	} else {
+		output.FatalfWithHintExamples([][]string{
+			{"Create new context with a sql container ", "sqlcmd create mssql"},
+		}, "Current context does not have a container")
+	}
```
shueybubbles, line 102: what will this look like for non-container connections? 
We can't claim this is production/enterprise ready until we support that.

I'd have expected most of the code from lines 76 to 96 to be outside the "if it's a container block"

Supporting a UNC or local file path to a bak file is a relatively straightforward restore operation

Restore from URL is also straightforward without copying the file locally if the server has a credential stored for the URL already.

cmd/modern/root/use.go
```diff
@@ -0,0 +1,107 @@
...
+			Mechanism: c.useMechanism,
+		})
+
+		if !useDatabase.SourceFileExists() {
+			output.FatalfWithHints(
+				[]string{fmt.Sprintf("File does not exist at URL %q", c.url)},
+				"Unable to download file to container")
+		}
+
+		useDatabase.CopyToContainer(id)
```
shueybubbles, line 86: >CopyToContainer(id) [](http://example.com/codeflow?start=14&length=19)

for .bak files using a URL we should provide a way to create a CREDENTIAL to store on the server so it can restore directly from the URL 

SQL 2022 supports both https and s3 URLs

cmd/modern/root/use.go
```diff
@@ -0,0 +1,107 @@
...
+
+		if useDatabase.IsExtractionNeeded() {
+			output.Infof("Extracting files from %q", useDatabase.UrlFilename())
+			useDatabase.Extract()
+		}
+
+		useDatabase.BringOnline(
+			c.sql.Query,
+			user.BasicAuth.Username,
+			secret.Decode(user.BasicAuth.Password, user.BasicAuth.PasswordEncryption),
```
shueybubbles, line 96: can we have a "credential" or "authentication" abstraction associated with the user to pass around instead of explicit user name and password? We're going to need to support all the auth types.

internal/container/controller.go
```diff
@@ -231,6 +233,40 @@ func (c Controller) ContainerFiles(id string, filespec string) (files []string)
...
+		panic("Must pass in non-empty destFolder")
+	}
+
+	_, f := filepath.Split(src)
+	h, err := os.ReadFile(src)
+	checkErr(err)
+
+	// Create and add some files to the archive.
+	var buf bytes.Buffer
+	tw := tar.NewWriter(&buf)
```
shueybubbles, line 253: >tw := tar.NewWriter(&buf) [](http://example.com/codeflow?start=1&length=25)

instead of having tw.Close below, shouldn't there be a "defer tw.Close()" call right after this? Nobody ever error checks a Close afaik

internal/container/controller.go
```diff
@@ -231,6 +233,40 @@ func (c Controller) ContainerFiles(id string, filespec string) (files []string)
...
+	}
+	err = tw.WriteHeader(hdr)
+	checkErr(err)
+	_, err = tw.Write([]byte(h))
+	checkErr(err)
+	err = tw.Close()
+	checkErr(err)
+
+	err = c.cli.CopyToContainer(context.Background(), id, destFolder, &buf, types.CopyToContainerOptions{})
+	checkErr(err)
```
shueybubbles, line 267: >kErr(err) [](http://example.com/codeflow?start=5&length=9)

do you need to delete the file after copying it or if there's an error?

internal/uri/type.go
```diff
@@ -0,0 +1,8 @@
+package uri
+
+import "net/url"
+
+type Uri struct {
+	uri string
```
shueybubbles, line 6: >uri string [](http://example.com/codeflow?start=1&length=10)

what's the benefit of keeping the string around? can't we get it from the URL?

internal/uri/uri.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+func (u Uri) IsLocal() bool {
+	if len(u.Scheme()) > 2 {
+		return false
+	} else {
```
shueybubbles, line 11: >else { [](http://example.com/codeflow?start=3&length=6)

no else needed here or on line 21

shueybubbles: couldn't this just be 
`return len(u.Scheme()) <= 2`

internal/uri/uri.go
```diff
@@ -0,0 +1,71 @@
...
+		return true
+	}
+}
+
+// actualUrl returns the url without the query string
+func (u Uri) ActualUrl() string {
+	terminator := strings.LastIndex(u.uri, ",")
```
shueybubbles, line 18: >, [](http://example.com/codeflow?start=41&length=1)

I think the comment is inaccurate. The query string is delimited by `?`

internal/uri/uri.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (u Uri) FileExtension() string {
+	_, f := filepath.Split(u.ActualUrl())
+	return strings.TrimLeft(filepath.Ext(f), ".")
+}
+
+func (u Uri) Filename() string {
+	filename := filepath.Base(u.ActualUrl())
+	if filename == "" {
+		panic("filename is empty")
```
shueybubbles, line 38: >	panic("filename is empty") [](http://example.com/codeflow?start=1&length=27)

this reliance on file name in the user's URL seems very limiting, even at this phase of our development. 
It's fine for the Uri to say "I don't know the file name" then let the consumer decide what to do. The consumer might say "ok let's download it and see what we get" or "fail if the user didn't tell us what mechanism to use".
Presumably if the mechanism is "restore" we can assume it's a "bak", correct?

internal/uri/uri.go
```diff
@@ -0,0 +1,71 @@
...
+func (u Uri) ParseDbName() string {
+	if u.uri == "" {
+		panic("uri is empty")
+	}
+
+	// TODO: Reimplement
+	return ""
+}
+
+func (u Uri) GetDbNameAsIdentifier() string {
```
shueybubbles, line 56: >etDbNameAsIdenti [](http://example.com/codeflow?start=14&length=16)

these SQL query escaping functions don't belong here.
They should be in their own package for general purpose use to safely inject identifiers into format strings.

internal/uri/uri_test.go
```diff
@@ -1,11 +1,64 @@
...
+}
+
+func TestParseDbName(t *testing.T) {
+	type test struct {
+		inputURL    string
+		expectedURL string
+	}
+
+	tests := []test{
+		{"https://example.com/testdb.bak,myDbName", "myDbName"},
```
shueybubbles, line 34: >		{"https://example.com/testdb.bak,myDbName" [](http://example.com/codeflow?start=0&length=44)

It's a bit odd to conflate "URL + a database name" as "Uri"

The fact that `--use` flag and "use" command take this format as a parameter is specific to the semantic of those commands, correct? 
I would expect these strings to be deserialized into something more meaningful like
```go
type DatabaseFile struct {
  Source *url.URL
  DatabaseName string
}
```

internal/uri/uri_test.go
```diff
@@ -1,11 +1,64 @@
...
+		{"https://example.com/testdb.bak", "testdb"},
+		{"https://example.com/test.foo", "test"},
+		{"https://example.com/test.foo,test", "test"},
+		{"https://example.com/test.7z,tql_name", "tsql_name"},
+		{"https://example.com/test.mdf,tsql_name?foo=bar", "tsql_name"},
```
shueybubbles, line 39: >"https://example.com/test.mdf,tsql_name?foo=bar [](http://example.com/codeflow?start=3&length=47)

how does putting the query portion of the URL after the comma work? It doesn't look like valid input.

A URL with a query would be `[https://example.com/test.mdf?version=10,mydatabase`](https://example.com/test.mdf?version=10,mydatabase`)

pkg/mssqlcontainer/ingest/ingest.go
```diff
@@ -0,0 +1,135 @@
...
+	query       func(text string)
+}
+
+func (i *ingest) IsExtractionNeeded() bool {
+	i.extractor = extract.NewExtractor(i.uri.FileExtension(), i.controller)
+	if i.extractor == nil {
+		return false
+	} else {
+		return true
+	}
```
shueybubbles, line 31: I propose  we pick one of three styles to avoid superfluous "else" blocks

1:
```go
return i.extractor != nil
```

2.
```go
func (i *ingest) IsExtractionNeeded() needed bool {
i.extractor = extract.NewExtractor(i.uri.FileExtension(), i.controller)
needed = iExtractor != nil
}
```

3. 
```go
func (i *ingest) IsExtractionNeeded() bool {
 i.extractor = extract.NewExtractor(i.uri.FileExtension(), i.controller)
 if i.extractor == nil {
  return false
 }
  return true
}
```

My vote is 1 .

pkg/mssqlcontainer/ingest/ingest.go
```diff
@@ -0,0 +1,135 @@
...
+	if i.location == nil {
+		panic("location is nil, did you call NewIngest()?")
+	}
+
+	i.containerId = containerId
+	i.location.CopyToContainer(containerId, destFolder)
+	i.options.Filename = i.uri.Filename()
+
+	if i.options.Filename == "" {
+		panic("filename is empty")
```
shueybubbles, line 65: >ename is empty") [](http://example.com/codeflow?start=12&length=16)

if the user provides a mechanism but the URL doesn't have a detectable file name, can we just generate a unique name with an extension that matches the mechanism?

pkg/mssqlcontainer/ingest/ingest.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (i *ingest) setDefaultDatabase(username string) {
+	if i.query == nil {
+		panic("query is nil, did you call BringOnline()?")
+	}
+
+	alterDefaultDb := fmt.Sprintf(
+		"ALTER LOGIN [%s] WITH DEFAULT_DATABASE = [%s]",
+		username,
```
shueybubbles, line 106: >username, [](http://example.com/codeflow?start=2&length=9)

user names need to be bracket escaped as well.
Fundamentally, we need a package with the same helper functions SMO has for safe scripting. See [https://github.com/microsoft/sqlmanagementobjects/blob/54e95cd095e232e43e711e65a83d55514ddc5e0e/src/Microsoft/SqlServer/Management/Smo/SqlSmoObject.cs#L329](https://github.com/microsoft/sqlmanagementobjects/blob/54e95cd095e232e43e711e65a83d55514ddc5e0e/src/Microsoft/SqlServer/Management/Smo/SqlSmoObject.cs#L329)

internal/uri/factory.go
```diff
@@ -0,0 +1,17 @@
+package uri
```
shueybubbles, line 1: add the copyright header to your new files

pkg/mssqlcontainer/ingest/extract/7zip.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (e *sevenZip) FileTypes() []string {
+	return []string{"7z"}
+}
+
+func (e *sevenZip) IsInstalled(containerId string) bool {
+	e.containerId = containerId
+
+	return false
```
shueybubbles, line 25: >return false [](http://example.com/codeflow?start=1&length=12)

tbd?

pkg/mssqlcontainer/ingest/extract/7zip.go
```diff
@@ -0,0 +1,95 @@
...
+func (e *sevenZip) Install() {
+	e.controller.RunCmdInContainer(e.containerId, []string{
+		"mkdir",
+		"/opt/7-zip"})
+
+	e.controller.RunCmdInContainer(e.containerId, []string{
+		"wget",
+		"-O",
+		"/opt/7-zip/7-zip.tar",
+		"https://7-zip.org/a/7z2201-linux-x64.tar.xz"})
```
shueybubbles, line 70: >x64 [](http://example.com/codeflow?start=36&length=3)

don't we need to get the architecture at runtime depending on where sqlcmd is running?

pkg/mssqlcontainer/ingest/extract/7zip.go
```diff
@@ -0,0 +1,95 @@
...
+	"path/filepath"
+	"regexp"
+)
+
+type sevenZip struct {
+	controller  *container.Controller
+	containerId string
+}
+
+func (e *sevenZip) Initialize(controller *container.Controller) {
```
shueybubbles, line 14: > Initialize(controller *container.Controller) [](http://example.com/codeflow?start=18&length=45)

How can we abstract away the differences between
"SQL running in a container", "SQL running on the local host",  "SQL running in the cloud", and "SQL running on a remote host"? 
Or would we have multiple extractor implementations and this one should be renamed "sevenZipInLinuxX64Container" and we'd have others like "windowsZipLocalHost" ?

I don't see how this model evolves to support the non-container scenarios.

shueybubbles:
Could you create a couple mermaid diagrams? One would show the relationship between components like `ingest` and `extract` and `container`, and one would show how the contents of the bak/ldf/dacpac file would move from their source to the database for each environment we support (sql in a container, sql on localhost, sql on a remote host, sql in the cloud).

```mermaid
a == label ==> b
```

pkg/mssqlcontainer/ingest/extract/7zip.go
```diff
@@ -0,0 +1,95 @@
...
+
+	e.controller.RunCmdInContainer(e.containerId, []string{
+		"chmod",
+		"u+x",
+		"/opt/7-zip/7zz",
+	})
+}
+
+func extractPaths(input string) []string {
+	re := regexp.MustCompile(`Path\s*=\s*(\S+)`)
```
shueybubbles, line 88: >`Path [](http://example.com/codeflow?start=26&length=5)

i assume this is never localized

pkg/mssqlcontainer/ingest/extract/7zip_test.go
```diff
@@ -0,0 +1,42 @@
...
+Modified = 2018-09-11 11:30:55.3152489
+Attributes = A
+CRC = BCA9F91F
+Encrypted = -
+Method = LZMA2:27
+Block = 2`
+
+	paths := extractPaths(stdout)
+
+	fmt.Println(paths)
```
shueybubbles, line 41: add some asserts

pkg/mssqlcontainer/ingest/location/factory.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/microsoft/go-sqlcmd/internal/container"
+)
+
+func NewLocation(isLocal bool, uri string, controller *container.Controller) Location {
+	if isLocal {
+		return local{
+			uri:        uri,
+			controller: controller,
+		}
+	} else {
```
shueybubbles, line 13: remove else

pkg/mssqlcontainer/ingest/mechanism/attach.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (m *attach) BringOnline(databaseName string, containerId string, query func(string), options BringOnlineOptions) {
+	text := `SET NOCOUNT ON; `
+
+	m.containerId = containerId
+	m.setFilePermissions(m.CopyToLocation() + "/" + options.Filename)
+	if options.LdfFilename == "" {
+		text += `CREATE DATABASE [%s] ON (FILENAME = '%s/%s') FOR ATTACH;`
+		query(fmt.Sprintf(
```
shueybubbles, line 36: >query(fmt.Sprintf [](http://example.com/codeflow?start=2&length=17)

you need to escape the parameters properly, and use nvarchar for the file name

eg from SMO
```C#
ddl.AppendFormat(SmoApplication.DefaultCulture, ", FILENAME = N'{0}' ", SqlString(fileName));

        public static String SqlString(String s)
        {
            return EscapeString(s, '\'');
        }
```

shueybubbles: it'd be a good test case to have some file names that use extended characters. Go stores strings as UTF8 but nvarchar is UTF16, while varchar is SQL-collation-dependent. 
How do we guarantee the Go strings get encoded properly in the query?

shueybubbles: maybe go-mssqldb does the right thing already for strings and we should use a parameterized query to let the driver do the conversion instead of manually inserting tokens in the query.

pkg/mssqlcontainer/ingest/mechanism/options.go
```diff
@@ -0,0 +1,9 @@
+package mechanism
+
+type BringOnlineOptions struct {
+	Username string
+	Password string
+
+	Filename    string
+	LdfFilename string
```
shueybubbles, line 8: >LdfFilename string [](http://example.com/codeflow?start=1&length=18)

should this be an array?

pkg/mssqlcontainer/ingest/mechanism/bacpac.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (m *bacpac) installSqlPackage() {
+	if m.controller == nil {
+		panic("controller is nil")
+	}
+
+	m.controller.DownloadFile(
+		m.containerId,
+		"https://aka.ms/sqlpackage-linux",
```
shueybubbles, line 62: >s/sqlpackage-linux", [](http://example.com/codeflow?start=16&length=20)

is sqlpackage CPU architecture agnostic? Does this also bring down Net Core runtime?

pkg/mssqlcontainer/ingest/mechanism/restore.go
```diff
@@ -0,0 +1,85 @@
...
+func (m *restore) BringOnline(databaseName string, _ string, query func(string), options BringOnlineOptions) {
+	if options.Filename == "" {
+		panic("Filename is required for restore")
+	}
+	if databaseName == "" {
+		panic("databaseName is required for restore")
+	}
+
+	query(fmt.Sprintf(
+		m.restoreStatement(),
```
shueybubbles, line 35: escape the params correctly. 
it's a bit trickier with dynamic sql.

shueybubbles (changes requested):
:clock1:

stuartpa:
> What about supporting non-public URLs (SAS etc)? Does it download the file then discover its type similar to how you'd have to examine a zip to know its content?

Doesn't the non-pupil SAS URL contain the filename still (before the ?), e.g.:

```
https://blahblah.blob.core.windows.net/database/AdventureWorks.bak?sv=2021-10-04&st=2023-04-07T06%3A12%3A15Z&se=2023-04-08T06%3A12%3A15Z&sr=b&sp=r&sig=<REDACTED>
```

Are there URL types that don't container the filename (e.g .bak or .7z?)

pkg/mssqlcontainer/ingest/mechanism/restore.go
```diff
@@ -0,0 +1,85 @@
...
+		m.CopyToLocation(),
+		options.Filename,
+		databaseName,
+		m.CopyToLocation(),
+		options.Filename,
+	))
+}
+
+func (m *restore) restoreStatement() string {
```
shueybubbles, line 44: >oreStatement() st [](http://example.com/codeflow?start=22&length=17)

this code is a good argument for porting SMO to AoT, as SMO already has all the backup and restore logic which has been tested for years.

shueybubbles:
A URL can be anything.  eg "aka.ms/ssmsfullsetup" is a pointer to an EXE. 



---
In reply to: [1499995103](https://github.com/microsoft/go-sqlcmd/pull/319#issuecomment-1499995103) [](http://example.com/codeflow?ancestors=1499995103,1499995103)

stuartpa:
This will work.  If the mechanism can't be determined from the file extension, then the --use-mechanism flag can be used (restore, dacfx, attach etc.).  I'll write unit tests around all these scenarios

---
In reply to: [1499460199](https://github.com/microsoft/go-sqlcmd/pull/319#issuecomment-1499460199) [](http://example.com/codeflow?ancestors=1499460199,1499460199)

stuartpa:
sure

---
In reply to: [1499673629](https://github.com/microsoft/go-sqlcmd/pull/319#issuecomment-1499673629) [](http://example.com/codeflow?ancestors=1499673629,1499673629)

stuartpa:
There is a --use-mechanism flag, that is required if we can't determine the mechanism from the file extension.

---
In reply to: [1500333712](https://github.com/microsoft/go-sqlcmd/pull/319#issuecomment-1500333712) [](http://example.com/codeflow?ancestors=1500333712,1499995103)

cmd/modern/root/use.go
stuartpa: This PR is only for container scenarios (I'll update the PR description). I am not sure if non-container scenarios are needed. (folks already have solutions for all of this).   Containers are particularly tricky in this regard for our normal users (who are windows centric).  This functionality abstracts most of the users during the early part of their journey from all these internals.

cmd/modern/root/use.go
stuartpa: Added an issue for this:

[https://github.com/microsoft/go-sqlcmd/issues/332](https://github.com/microsoft/go-sqlcmd/issues/332)

PR #340, opened 2023-05-16
Do not read basic auth info if auth type is other

### LINKED ISSUE #339: BasicAuthentication assumed by default if user is not nil
Isolating this issue from #329 
Refer #329 

Segfault
I have this segfault. I solved it by deleting sqlcmd's config file.
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x81990d]

goroutine 1 [running]:
github.com/microsoft/go-sqlcmd/internal/config.GetCurrentContextInfo()
       /home/vsts/work/1/s/internal/config/context.go:105 +0x10d
main.initializeEnvVars()
       /home/vsts/work/1/s/cmd/modern/main.go:83 +0xcf
main.main()
       /home/vsts/work/1/s/cmd/modern/main.go:51 +0x20c
```

apoorvdeshmukh:
This commit fixes #339 
`sqlcmd config add-user` command by default assumes basic authentication type but in case a user is created with auth type other than "basic" then connection will be established to endpoint mentioned by current-context (which is existing behavior) and further auth info won't be read as auth type "other" is not supported yet.

### DIFF
--- a/internal/config/context.go
+++ b/internal/config/context.go
@@ -101,7 +101,7 @@ func CurrentContext() (endpoint Endpoint, user *User) {
 func GetCurrentContextInfo() (server string, username string, password string) {
 	endpoint, user := CurrentContext()
 	server = fmt.Sprintf("%s,%d", endpoint.Address, endpoint.Port)
-	if user != nil {
+	if user != nil && user.AuthenticationType == "basic" {
 		username = user.BasicAuth.Username
 		if user.AuthenticationType == "basic" {
 			password = decryptCallback(


internal/config/context.go
```diff
@@ -101,7 +101,7 @@ func CurrentContext() (endpoint Endpoint, user *User) {
 func GetCurrentContextInfo() (server string, username string, password string) {
 	endpoint, user := CurrentContext()
 	server = fmt.Sprintf("%s,%d", endpoint.Address, endpoint.Port)
-	if user != nil {
+	if user != nil && user.AuthenticationType == "basic" {
 		username = user.BasicAuth.Username
 		if user.AuthenticationType == "basic" {
```
shueybubbles, line 106: >	if user.AuthenticationType == "basic" { [](http://example.com/codeflow?start=1&length=40)

this line is redundant. 
It also implies that there could be a case where we still need a user name even when authenticationType is not "basic". 

What will the CurrentContext return for authentication types like AAD interactive that have a user name but no password?

shueybubbles (approved):
:shipit:

internal/config/context.go
apoorvdeshmukh: In main() this is only used if sqlcmd is invoked without any parameters and current-context is set in config file.
In that case, it is possible that the current-context may have a user associated with that context with AuthnticationType "Other".
`sqlcmd config add-user` generally creates user with 'Basic' auth-type as default but one could explicitly set this type as 'other'.

Issue #342, opened 2023-05-17
Cobra Migration - screen width is appearing twice


JyotikaGargg:
There is something weird going on with -w:
Help shows :
  -w, --w int                           Specifies the screen width for output
And I also see:
      --screen-width int                Specifies the screen width for output
      
It should be single instance.

PR #358, opened 2023-05-25
removing extra screenwidth entry

JyotikaGargg:
Related issue : 
https://github.com/microsoft/go-sqlcmd/issues/342

### DIFF
--- a/cmd/sqlcmd/sqlcmd.go
+++ b/cmd/sqlcmd/sqlcmd.go
@@ -235,7 +235,7 @@ func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
 
 	var screenWidth int
 	screenWidthPtr := &screenWidth
-	rootCmd.Flags().IntVarP(screenWidthPtr, "screen-width", "w", 0, "Specifies the screen width for output")
+	rootCmd.Flags().IntVarP(screenWidthPtr, "screen-width", "w", 0, localizer.Sprintf("Specifies the screen width for output"))
 }
 
 func normalizeFlags(rootCmd *cobra.Command) error {
--- a/cmd/sqlcmd/sqlcmd.go
+++ b/cmd/sqlcmd/sqlcmd.go
@@ -232,8 +232,10 @@ func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
 
 	// Using PersistentFlags() for ErrorSeverityLevel due to data type uint8 , which is not supported in Flags()
 	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "error-severity-level", "V", 0, "Controls the severity level that is used to set the ERRORLEVEL variable on exit.")
-	screenWidth := rootCmd.Flags().Int("screen-width", 0, "Specifies the screen width for output")
-	rootCmd.Flags().IntVarP(screenWidth, "w", "w", 0, "Specifies the screen width for output")
+
+	var screenWidth int
+	screenWidthPtr := &screenWidth
+	rootCmd.Flags().IntVarP(screenWidthPtr, "screen-width", "w", 0, "Specifies the screen width for output")
 }
 
 func normalizeFlags(rootCmd *cobra.Command) error {


shueybubbles (approved):
:shipit:

cmd/sqlcmd/sqlcmd.go
```diff
@@ -232,8 +232,10 @@ func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
 
 	// Using PersistentFlags() for ErrorSeverityLevel due to data type uint8 , which is not supported in Flags()
 	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "error-severity-level", "V", 0, "Controls the severity level that is used to set the ERRORLEVEL variable on exit.")
-	screenWidth := rootCmd.Flags().Int("screen-width", 0, "Specifies the screen width for output")
-	rootCmd.Flags().IntVarP(screenWidth, "w", "w", 0, "Specifies the screen width for output")
+
+	var screenWidth int
+	screenWidthPtr := &screenWidth
+	rootCmd.Flags().IntVarP(screenWidthPtr, "screen-width", "w", 0, "Specifies the screen width for output")
```
shueybubbles, line 238: >Specifies the sc [](http://example.com/codeflow?start=66&length=16)

localizer

shueybubbles:
I think part of the cobra conversion has caused `go test cmd/sqlcmd` to print a lot of stuff to the screen now. Is there a way to redirect that output in the tests?

shueybubbles:
I tried changing the test to set `SilenceErrors` and `SilenceUsage`

```go
  cmd := &cobra.Command{
   Use:   "testCommand",
   Short: "A brief description of my command",
   Long:  "A long description of my command",
   PreRunE: func(cmd *cobra.Command, argss []string) error {
    SetScreenWidthFlag(arguments, cmd)
    return arguments.Validate()
   },
   Run: func(cmd *cobra.Command, argss []string) {
    // Command logic goes here
   },
   SilenceErrors: true,
   SilenceUsage:  true,
  }

```
but I still get this:

```
Running tool: C:\Program Files\Go\bin\go.exe test -timeout 90s -tags np,sm -run ^TestValidCommandLineToArgsConversion$ github.com/microsoft/go-sqlcmd/cmd/sqlcmd -v

=== RUN   TestValidCommandLineToArgsConversion
A long description of my command

Usage:
  testCommand [flags]

Flags:
  -K, --application-intent string       Declares the application workload type when connecting to a server. The only currently supported value is ReadOnly. If -K is not specified, the sqlcmd utility will not support connectivity to a secondary replica in an Always On availability group (default "default")
      --authentication-method string    Specifies the SQL authentication method to use to connect to Azure SQL Database. One of:ActiveDirectoryDefault,ActiveDirectoryIntegrated,ActiveDirectoryPassword,ActiveDirectoryInteractive,ActiveDirectoryManagedIdentity,ActiveDirectoryServicePrincipal,SqlPassword
  -c, --batch-terminator string         Specifies the batch terminator. The default value is GO (default "GO")
  -s, --column-separator string         Specifies the column separator character. Sets the SQLCMDCOLSEP variable.
  -d, --database-name string            This option sets the sqlcmd scripting variable SQLCMDDBNAME. This parameter specifies the initial database. The default is your login's default-database property. If the database does not exist, an error message is generated and sqlcmd exits.
  -A, --dedicated-admin-connection      Dedicated administrator connection
  -X, --disable-cmd-and-warn            Disables commands that might compromise system security. Sqlcmd issues a warning and continues.
  -x, --disable-variable-substitution   Causes sqlcmd to ignore scripting variables. This parameter is useful when a script contains many INSERT statements that may contain strings that have the same format as regular variables, such as $(variable_name).
      --driver-logging-level int        Level of mssql driver messages to print.
  -N, --encrypt-connection string       This switch is used by the client to request an encrypted connection. (default "default")
  -m, --error-level int                 Controls which error messages are sent to stdout. Messages that have severity level greater than or equal to this level are sent.
  -V, --error-severity-level uint8      Controls the severity level that is used to set the ERRORLEVEL variable on exit.
  -r, --errors-to-stderr int            Controls which error messages are sent to stdout. Messages that have severity level greater than or equal to this level are sent. (default -1)
  -b, --exit-on-error                   Specifies that sqlcmd exits and returns a DOS ERRORLEVEL value when an error occurs.
  -F, --format string                   Specifies the formatting for results. (default "horiz")
  -h, --headers int                     Specifies the number of rows to print between the column headings. Use -h-1 to specify that headers not be printed.
  -?, --help                            -? shows this syntax summary, --help shows modern sqlcmd sub-command help
  -q, --initial-query string            Executes a query when sqlcmd starts, but does not exit sqlcmd when the query has finished running. Multiple-semicolon-delimited queries can be executed.
  -i, --input-file strings              Identifies one or more files that contain batches of SQL statements. If one or more files do not exist, sqlcmd will exit. Mutually exclusive with -Q/-q.
  -l, --login-timeOut int               Specifies the number of seconds before a sqlcmd login to the go-mssqldb driver times out when you try to connect to a server. This option sets the sqlcmd scripting variable SQLCMDLOGINTIMEOUT. The default value is 30. 0 means infinite. (default -1)
  -M, --multi-subnet-failover           Provided for backward compatibility. Sqlcmd always optimizes detection of the active replica of a SQL Failover Cluster.
  -o, --output-file string              Identifies the file that receives output from sqlcmd.
  -a, --packet-size int                 Requests a packet of a different size. This option sets the sqlcmd scripting variable SQLCMDPACKETSIZE. packet_size must be a value between 512 and 32767. The default = 4096. A larger packet size can enhance performance for execution of scripts that have lots of SQL statements between GO commands. You can request a larger packet size. However, if the request is denied, sqlcmd uses the server default for packet size.
  -P, --password string                 Obsolete. The initial passwords must be set using the SQLCMDPASSWORD environment variable or entered at the password prompt.
  -Q, --query string                    Executes a query when sqlcmd starts and then immediately exits sqlcmd. Multiple-semicolon-delimited queries can be executed.
      --screen-width int                Specifies the screen width for output
  -S, --server string                   [[tcp:]|[lpc:]|[np:]]server[\instance_name][,port]Specifies the instance of SQL Server to which to connect. It sets the sqlcmd scripting variable SQLCMDSERVER.
  -W, --trim-spaces                     Remove trailing spaces from a column.
  -C, --trust-server-certificate        Implicitly trust the server certificate without validation.
  -u, --unicode-output-file             Specifies that all output files are encoded with little-endian Unicode
  -G, --use-aad                         Tells sqlcmd to use Active Directory authentication. If no user name is provided, authentication method ActiveDirectoryDefault is used. If a password is provided, ActiveDirectoryPassword is used. Otherwise ActiveDirectoryInteractive is used.
  -E, --use-trusted-connection          Uses a trusted connection instead of using a user name and password to sign in to SQL Server, ignoring any environment variables that define user name and password.
  -U, --user-name string                The login name or contained database user name.  For contained database users, you must provide the database name option
  -v, --variables stringToString        Creates a sqlcmd scripting variable that can be used in a sqlcmd script. Enclose the value in quotation marks if the value contains spaces. You can specify multiple var=values values. If there are errors in any of the values specified, sqlcmd generates an error message and then exits (default [])
      --version                         Print version information and exit
  -w, --w int                           Specifies the screen width for output
  -H, --workstation-name string         This option sets the sqlcmd scripting variable SQLCMDWORKSTATION. The workstation name is listed in the hostname column of the sys.sysprocesses catalog view and can be returned using the stored procedure sp_who. If this option is not specified, the default is the current computer name. This name can be used to identify different sqlcmd sessions.
--- PASS: TestValidCommandLineToArgsConversion (0.24s)
PASS
ok   github.com/microsoft/go-sqlcmd/cmd/sqlcmd 0.615s
```

---
In reply to: [1563391313](https://github.com/microsoft/go-sqlcmd/pull/358#issuecomment-1563391313) [](http://example.com/codeflow?ancestors=1563391313)

shueybubbles:
oh nm this is because we're testing with -h, which will always print the help during Execute.
I will change the command's output pipe instead.

---
In reply to: [1563409319](https://github.com/microsoft/go-sqlcmd/pull/358#issuecomment-1563409319) [](http://example.com/codeflow?ancestors=1563409319,1563391313)

stuartpa (approved):
:shipit:

PR #369, opened 2023-06-03
feat:Add -P -y -Y flags

### LINKED ISSUE #196: Restore `-P` for backward compatibility.
`go-sqlcmd` works really well as a drop in replacement for the legacy sql server command line tools in most cases.  However, some 3rd party utilities, e.g. PeopleSoft ChangeAssistant, that rely on `-P` flag no longer work.  Would it be possible to restore the clear text password passing using the `-P` flag so that go-sqlcmd utility maintains the compatibility with its heavyweight sqlcmd cousin?

### LINKED ISSUE #362: sqlcmd shouldn't show the full set of flags when the user doesn't provide an argument parameter
EG ODBC

```

16:30:56.64 e:\git>sqlcmd -w
Sqlcmd: '-w': Missing argument. Enter '-?' for help.

```
And go-sqlcmd:

```
16:59:19.21 e:\git\go-sqlcmd>.\sqlcmd -w
Error: flag needs an argument: 'w' in -w
Usage:
   [flags]
<followed by every flag of which there are dozens>

```
The problem is it ends up scrolling the command prompt so far that I can't see what parameter I didn't provide.

Can we switch to only showing the full set of flags when the user passes `-?` , and giving a reminder to use `-?` when an argument is missing?

### LINKED ISSUE #365: Implement -y and -Y

shueybubbles:
Fixes #365 
Fixes #196 
Fixes #362 
I've also standardized more of the invalid flag error messages and replaced the default Cobra errors with localizable errors prefixed with `Sqlcmd: `

### DIFF
--- a/cmd/sqlcmd/sqlcmd.go
+++ b/cmd/sqlcmd/sqlcmd.go
@@ -292,11 +292,14 @@ func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
 	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "error-severity-level", "V", 0, localizer.Sprintf("Controls the severity level that is used to set the %s variable on exit", localizer.ErrorLevel))
 
 	_ = rootCmd.Flags().IntP(screenWidth, "w", 0, localizer.Sprintf("Specifies the screen width for output"))
-	_ = rootCmd.Flags().IntP(variableTypeWidth, "y", 256, localizer.Sprintf("Sets the sqlcmd scripting variable %s", "SQLCMDMAXVARTYPEWIDTH"))
-	_ = rootCmd.Flags().IntP(fixedTypeWidth, "Y", 0, localizer.Sprintf("Sets the sqlcmd scripting variable %s", "SQLCMDMAXFIXEDTYPEWIDTH"))
+	_ = rootCmd.Flags().IntP(variableTypeWidth, "y", 256, setScriptVariable("SQLCMDMAXVARTYPEWIDTH"))
+	_ = rootCmd.Flags().IntP(fixedTypeWidth, "Y", 0, setScriptVariable("SQLCMDMAXFIXEDTYPEWIDTH"))
 	rootCmd.Flags().BoolVarP(&args.DedicatedAdminConnection, "dedicated-admin-connection", "A", false, localizer.Sprintf("Dedicated administrator connection"))
 }
 
+func setScriptVariable(v string) string {
+	return localizer.Sprintf("Sets the sqlcmd scripting variable %s", v)
+}
 func normalizeFlags(cmd *cobra.Command) error {
 	//Adding a validator for checking the enum flags
 	var err error
--- a/cmd/sqlcmd/sqlcmd.go
+++ b/cmd/sqlcmd/sqlcmd.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -60,6 +61,8 @@ type SQLCmdArguments struct {
 	Version                     bool
 	ColumnSeparator             string
 	ScreenWidth                 *int
+	VariableTypeWidth           *int
+	FixedTypeWidth              *int
 	TrimSpaces                  bool
 	MultiSubnetFailover         bool
 	Password                    string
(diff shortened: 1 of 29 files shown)


cmd/sqlcmd/sqlcmd.go
```diff
@@ -262,66 +286,127 @@ func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
...
 
 	// Using PersistentFlags() for ErrorSeverityLevel due to data type uint8 , which is not supported in Flags()
 	rootCmd.PersistentFlags().Uint8VarP(&args.ErrorSeverityLevel, "error-severity-level", "V", 0, localizer.Sprintf("Controls the severity level that is used to set the %s variable on exit", localizer.ErrorLevel))
 
-	var screenWidth int
-	screenWidthPtr := &screenWidth
-	rootCmd.Flags().IntVarP(screenWidthPtr, "screen-width", "w", 0, localizer.Sprintf("Specifies the screen width for output"))
+	_ = rootCmd.Flags().IntP(screenWidth, "w", 0, localizer.Sprintf("Specifies the screen width for output"))
+	_ = rootCmd.Flags().IntP(variableTypeWidth, "y", 256, localizer.Sprintf("Sets the sqlcmd scripting variable %s", "SQLCMDMAXVARTYPEWIDTH"))
+	_ = rootCmd.Flags().IntP(fixedTypeWidth, "Y", 0, localizer.Sprintf("Sets the sqlcmd scripting variable %s", "SQLCMDMAXFIXEDTYPEWIDTH"))
```
apoorvdeshmukh, line 296: Can you check if you are getting any errors when generating localizations JSONs by running `go generate .\internal\translations\translations.go`
The gotext tool uses the string itself as a key when creating the map and if two keys are same then it simply throws warning and ignores the conflicting strings in resulting JSON file.

Following is the error for the above string
```
git\go-sqlcmd>go generate .\internal\translations\translations.go
Key "Sets the sqlcmd scripting variable %s" matches conflicting messages: [Sets the sqlcmd scripting variable {SQLCMDMAXVARTYPEWIDTH}] and [Sets the sqlcmd scripting variable {SQLCMDMAXFIXEDTYPEWIDTH}]
```

apoorvdeshmukh (changes requested):
You may want to commit the updated JSON files after running `go generate .\internal\translations\translations.go`

cmd/sqlcmd/sqlcmd.go
shueybubbles: is there a way to tell gotext to suppress the warnings about missing translations? It'd be ideal for `build\build` to only print the errors that can be fixed by code.

shueybubbles:
@apoorvdeshmukh  How do I fix this test?

```go
  {
   name:     "existing key",
   key:      "'-w %d': value must be greater than 8 and less than 65536.",
   expected: "k",
   ok:       true,
  },
```

cmd/sqlcmd/sqlcmd.go
```diff
@@ -262,66 +286,130 @@ func setFlags(rootCmd *cobra.Command, args *SQLCmdArguments) {
...
 		return pflag.NormalizedName(name)
 	})
+	if err != nil {
+		cmd.PrintErrln(sqlcmdErrorPrefix, err.Error())
+		cmd.SilenceErrors = true
+	}
 	return err
 }
+
+var invalidArgRegexp = regexp.MustCompile(`invalid argument \"(.*)\" for \"(-.), (--.*)\"`)
```
stuartpa, line 357: I was testing to see if these strings should be localizer-izsed... but loc seems to be broken in this branch:

set SQLCMD_LANG=de-de
.\sqlcmd.exe -w 4

Is returning English:

Sqlcmd: '-w 4': value must be greater than 8 and less than 65536.

It should be:

sqlcmd: Error: '-w 4': Der Wert muss größer als 8 und kleiner als 65536 sein.

stuartpa (changes requested):
:clock1:

cmd/sqlcmd/sqlcmd.go
apoorvdeshmukh: Not that I know of.

cmd/sqlcmd/sqlcmd.go
shueybubbles: I posted a discussion in Teams about standardizing more of the error messages.

apoorvdeshmukh: I had a look at the test which is failing. 
After running the `go generate` command, the catalog.go gets built and the map is generated again.
The key which the testcase is trying to look for doesn't exist in the map and hence it is failing.
I'll push another commit which fixes the testcase.

apoorvdeshmukh:
> @apoorvdeshmukh How do I fix this test?
> 
> ```go
>   {
>    name:     "existing key",
>    key:      "'-w %d': value must be greater than 8 and less than 65536.",
>    expected: "k",
>    ok:       true,
>   },
> ```
> 
> In reply to: [1579265257](https://github.com/microsoft/go-sqlcmd/pull/369#issuecomment-1579265257)

I had a look at the test which is failing. 
After running the `go generate` command, the catalog.go gets built and the map is generated again.
The key which the testcase is trying to look for doesn't exist in the map and hence it is failing.
It looks for existing key and then looks for it's index in the dictionary and uses the corresponding data as expected value. 
This testcase may need modifications if the index of the key  gets changed.

apoorvdeshmukh (approved):

PR #379, opened 2023-06-12
Update localization settings

apoorvdeshmukh:
This commit update the language IDs so that the localization artifacts are generated 
in same directories as the ones generated by Golang localization framework.
The output path is intentionally kept different to make sure that the initial pipeline
run does not affect the existing builds which contains partial localized data.
The localization team will checkin the localized artifacts in the output path.
We will keep the content received from localization team and the one used by our builds 
in sync via script.


apoorvdeshmukh:
After this PR is merged, localization pipeline will be run on main branch and the generated artifacts will be used for onboarding

shueybubbles (approved):
:shipit:

PR #382, opened 2023-06-13
Localized file check-in by OneLocBuild Task: Build definition ID 615:…

apoorvdeshmukh:
… Build ID 202973 (#381)

As the localization pipeline is failing as it cannot auto merge on main branch, creating this branch for use with localization pipeline.


shueybubbles:
should we give their account auto-merge privileges instead?

apoorvdeshmukh:
> should we give their account auto-merge privileges instead?

Yes. That would be great. Then we can run the pipeline on 'main' itself.

shueybubbles (approved):

PR #397, opened 2023-06-21
Fix: Miscellaneous ODBC compatibility fixes

### LINKED ISSUE #394: Restore -I as a no-op
go-sqlcmd connects with SET QUOTED_IDENTIFIER ON which is a good thing. This removes the need of the -I option that the old SQLCMD has, and therefore this option has been dropped.

However, I don't think this was a wise move. Say that you have a BAT file that loads SQL objects with SQLCMD. This BAT file may execute on multiple machines, and you have no control over which SQLCMD that is installed on the individual machine (or which comes first in the path).

If you leave -I on the SQLCMD command line, the script will fail on computers with go-sqlcmd.

If you take out -I, the script will load SQL objects with QUJOTED_IDENTIFIER OFF on machines with the old SQLCMD. This can leads to all sorts of problem.

Therefore, go-sqlcmd should accept -I as an option, but it should not affect anything. But just not cause an error.

### LINKED ISSUE #395: go-sqlcmd reports incorrect line numbers (because it strips blank lines)
[slask.txt](https://github.com/microsoft/go-sqlcmd/files/11824145/slask.txt)
If you attempt to load the attached file in the Northwind database, you will get this error:

```
Changed database context to 'Northwind'.
Msg 102, Level 15, State 1, Server SERVERNAME, Line 34
Incorrect syntax near '='.
```
If you account for the initial batch, you add 2 to 34 and you to line 36 in file to look at:
```
      JOIN   Products P ON P.ProductID = OD.ProductID
```
and you ask yourself, what's wrong with that?

Answer: absolutely nothing. But if you move to line 41 in the file, you find:
```
   IF @debug == 1
```
And that is not legal T-SQL.

If you remove the extraneous = and load the file again, it succeeds. If you run

```
sp_helptext ProductCountrSales_sp
```
You find why the line number was not reported correctly; All blank lines have been stripped out.

Lest you think this is a very small thing, it is not. This is a complete show-stopper for me to start using go-sqlcmd. I use SQLCMD to load stored procedures from my Editor, and having correct line numbers is absolutely essential.

### LINKED ISSUE #396: go-sqlcmd does not report procedure name
[slask.txt](https://github.com/microsoft/go-sqlcmd/files/11824153/slask.txt)
If you attempt to load the attached file in the Northwind database with go-sqlcmd, you get this output:
```
Changed database context to 'Northwind'.
Msg 102, Level 15, State 1, Server SERVERNAME, Line 34
Incorrect syntax near '='.
```
If you instead attempt to load it with old SQLCMD, you see this:
```
Changed database context to 'Northwind'.
Msg 102, Level 15, State 1, Server SERVERNAME, Procedure ProductCountrSales_sp, Line 39
Incorrect syntax near '='.
```
The difference is that the message from old SQLCMD includes the name of the stored procedure, but this is missing from go-sqlcmd. In this particular case, this is not a big deal, but imagine that if the file includes many stored procedures and other objects. Not knowing in which procedure the syntax error is hiding, is certainly problematic.

shueybubbles:
Fixes #394 
Fixes #395 
Fixes #396 
- Preserve blank lines in the batch
- Add procedure name to errors when appropriate
- Stub out -I and -R flags so they are allowed in scripts but have no effect on execution


apoorvdeshmukh (approved):
:shipit:

PR #409, opened 2023-07-06
[Draft] App insights 

JyotikaGargg:


build/build.cmd
```diff
@@ -26,7 +26,7 @@ if not exist %gopath%\bin\gotext.exe (
 )
 
 REM go-winres likes to append instead of overwrite so delete existing resource file
-del %~dp0..\cmd\modern\*.syso
```
shueybubbles, line 29: why remove this?

cmd/modern/main.go
```diff
@@ -38,19 +39,26 @@ var version = "local-build" // overridden in pipeline builds with: -ldflags="-X
...
 		cmdparser.Initialize(initializeCallback)
+		telemetry.TrackEvent("modern")
 		rootCmd.Execute()
 	} else {
 		initializeEnvVars()
+		telemetry.TrackEvent("legacy")
 		legacyCmd.Execute(version)
 	}
+	telemetry.FlushTelemetry()
+	telemetry.CloseTelemetry()
```
shueybubbles, line 61: move these up to line 43 and use defer
```Go
defer telemetry.CloseTelemetry()
defer telemetry.FlushTelemetry()
```

internal/telemetry/appinsights.go
```diff
@@ -0,0 +1,67 @@
...
+	// Add a diagnostics listener for printing telemetry messages
+	appinsights.NewDiagnosticsMessageListener(func(msg string) error {
+		fmt.Printf("[%s] %s\n", time.Now().Format(time.UnixDate), msg)
+		return nil
+	})
+}
+
+// TrackCommand tracks a command execution event
+func TrackCommand(command string) {
+	event := appinsights.NewEventTelemetry("command")
```
shueybubbles, line 38: >etry("command") [](http://example.com/codeflow?start=35&length=15)

Do we need a hierarchy/categorized name for events and properties so we can group things?

Eg in SSMS there are multiple events that happen to an object explorer node. The event names are `/` delimited like:
```
sql/ssms/explorerhierarchynode/setupmenuparent
                                    DataModel.Action.DurationInMilliseconds = 0
                                    SQL.SSMS.ObjectExplorer.Entity = Server/Login
                                    SQL.SSMS.ObjectExplorer.IsSqlDW = False
                                    SQL.SSMS.ObjectExplorer.ServerType = SqlAzureDatabase
sql/ssms/explorerhierarchynode/getmenuitems
                                    DataModel.Action.DurationInMilliseconds = 567.6678
                                    SQL.SSMS.ObjectExplorer.Entity = Server/Login
                                    SQL.SSMS.ObjectExplorer.IsSqlDW = False
                                    SQL.SSMS.ObjectExplorer.ServerType = SqlAzureDatabase
```
There are also a bunch of "ambient" properties we tried to include in most events, like the type of server OE is connected to.

In our case, I feel like `command` and `sub-command` don't need to be different events.
We can have a single `command` event that has a property like `CommandId` whose value is `<commandname>/<subcommand name>` 
When we are querying the backend data, we can more easily filter based on the `CommandId` property of one event instead of searching for separate events and manually aggregating them.

Also - for "modern" commands we should log the context id (a guid) with every relevant event.  That will enable us to track the lifecycle of a context. EG if the customer runs sqlcmd to create a context, there should be an event that logs the context id guid along with non-PII metadata about the context (container type, operating system of the host, version of SQL, etc). Subsequent commands using that context can then be correlated with that creation.

If the user uses an existing context for an operation, we can log a 'ConnectContext' event that has the similar metadata as the `CreateContext` event.

internal/telemetry/appinsights.go
```diff
@@ -0,0 +1,67 @@
...
+
+// TrackSubCommand tracks a sub-command execution event
+func TrackSubCommand(command, subCommand string) {
+	event := appinsights.NewEventTelemetry("sub-command")
+	event.Properties["command"] = command
+	event.Properties["sub-command"] = subCommand
+	telemetryClient.Track(event)
+}
+
+func TrackEvent(eventName string) {
```
shueybubbles, line 51: >TrackEvent [](http://example.com/codeflow?start=5&length=10)

this is going to need either a `map[string]string` or `[]string...` argument so callers can pass custom property values to the event.

map is probably more idiomatic, but some people like string arrays where you pass name and value separately like

`TrackEvent("myevent", "prop1Name", "prop1Value", "prop2Name", "prop2Value')`

internal/telemetry/appinsights.go
```diff
@@ -0,0 +1,67 @@
...
+	event := appinsights.NewEventTelemetry(eventName)
+	event.Properties["command"] = eventName
```
shueybubbles, line 53: >.Properties [](http://example.com/codeflow?start=6&length=11)

what kind of map is `Properties`? Can a value be an arbitrary Go object that implements a particular interface for serialization, or does it just take string values?

internal/telemetry/appinsights.go
```diff
@@ -0,0 +1,67 @@
+package telemetry
```
shueybubbles, line 1: >ackage telemetry [](http://example.com/codeflow?start=1&length=16)

it could be useful to define a method that accepts event name and properties plus a func() as input. It would run the function with a timer and add an elapsed time property to the telemetry event before posting it.

Something like
```
TrackTimedOperation(name string, properties map[string]string, func ())
```
then we can measure how long it takes to deploy things without trying to look for a "begin" event and its corresponding "end" event.

shueybubbles:
These basic method definitions are good for a POC.
It needs to be fleshed out, such as defining an interface for the internal telemetry implementation and using the dependency inversion infrastructure to pass the instance of the interface to the command implementations like we pass other services. We don't want to have all the commands calling a global singleton directly. Tests can provide a mock implementation of that interface to verify the commands log the expected telemetry when run.

internal/telemetry/appinsights.go
JyotikaGargg: This is of type string only. This is map of string keys and string values. We need to convert Go object as a value to the properties map by converting into string only .

internal/telemetry/appinsights.go
JyotikaGargg: This is a good suggestion , but i would like to add one POV here . Instead of every granular event to have start and end time , we should have these func parmater with specified events (such as deploy and create).

internal/telemetry/appinsights.go
JyotikaGargg: map[string]string is ideal to have . Changes done accordingly.

internal/telemetry/appinsights.go
shueybubbles: Yes I don't think every operation needs an elapsed time property, just certain ones. At the lowest level, though, the function just needs to take an event name and properties. The application-level code will decide which events need to use it.

PR #414, opened 2023-07-06
Fix incorrectly wrapped strings by localization routines

### LINKED ISSUE #413: sqlcmd prints incorrectly strings
The latest build prints erroneous strings.
```
>.\sqlcmd.exe stop
Stopping %!!(string=mcr.microsoft.com/mssql/server:latest)q(BADINDEX) for context %!!(string=mssql)q(BADINDEX)
```

apoorvdeshmukh:
Fixes #413 

```
git\go-sqlcmd>.\sqlcmd.exe start
Starting "mcr.microsoft.com/mssql/server:latest" for context "mssql"

git\go-sqlcmd>.\sqlcmd.exe stop
Stopping "mcr.microsoft.com/mssql/server:latest" for context "mssql"

git\go-sqlcmd>.\sqlcmd.exe remove --force --yes
Removing context mssql
Container "4909e2af7b209c3dd395911747882d90a90ee43ed7ed0ea335fdeb96481abd29" no longer exists, continuing to remove context...
Operation completed successfully
```

### DIFF
--- a/cmd/modern/root/install/mssql-base.go
+++ b/cmd/modern/root/install/mssql-base.go
@@ -290,7 +290,7 @@ func (c *MssqlBase) createContainer(imageName string, contextName string) {
 		c.downloadImage(imageName, output, controller)
 	}
 
-	output.Infof(localizer.Sprintf("Starting %v", imageName))
+	output.Info(localizer.Sprintf("Starting %v", imageName))
 	containerId := controller.ContainerRun(
 		imageName,
 		env,
@@ -319,7 +319,7 @@ func (c *MssqlBase) createContainer(imageName string, contextName string) {
 		c.passwordEncryption,
 	)
 
-	output.Infof(
+	output.Info(
 		localizer.Sprintf("Created context %q in \"%s\", configuring user account...",
 			config.CurrentContextName(),
 			config.GetConfigFileUsed()))
@@ -327,7 +327,7 @@ func (c *MssqlBase) createContainer(imageName string, contextName string) {
 	controller.ContainerWaitForLogEntry(
 		containerId, c.errorLogEntryToWaitFor)
 
-	output.Infof(
+	output.Info(
 		localizer.Sprintf("Disabled %q account (and rotated %q password). Creating user %q",
 			"sa",
 			"sa",
@@ -452,7 +452,7 @@ func (c *MssqlBase) createNonSaUser(
 		defaultDatabase = c.defaultDatabase
 
 		// Create the default database, if it isn't a downloaded database
-		output.Infof(localizer.Sprintf("Creating default database [%s]", defaultDatabase))
+		output.Info(localizer.Sprintf("Creating default database [%s]", defaultDatabase))
 		c.query(fmt.Sprintf("CREATE DATABASE [%s]", defaultDatabase))
 	}
 
@@ -532,7 +532,7 @@ func (c *MssqlBase) downloadAndRestoreDb(
 	_, file := filepath.Split(databaseUrl)
 
 	// Download file from URL into container
-	output.Infof(localizer.Sprintf("Downloading %s", file))
+	output.Info(localizer.Sprintf("Downloading %s", file))
 
 	temporaryFolder := "/var/opt/mssql/backup"
 
@@ -543,7 +543,7 @@ func (c *MssqlBase) downloadAndRestoreDb(
 	)
 
 	// Restore database from file
-	output.Infof(localizer.Sprintf("Restoring database %s", databaseName))
+	output.Info(localizer.Sprintf("Restoring database %s", databaseName))
 
 	dbNameAsIdentifier := getDbNameAsIdentifier(databaseName)
 	dbNameAsNonIdentifier := getDbNameAsNonIdentifier(databaseName)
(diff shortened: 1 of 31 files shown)


cmd/modern/root/uninstall.go
```diff
@@ -116,10 +116,7 @@ func (c *Uninstall) run() {
 			output.Infof(localizer.Sprintf("Removing context %s", config.CurrentContextName()))
 
 			if controller.ContainerExists(id) {
-				output.Infof(
-					"Stopping %s",
-					endpoint.ContainerDetails.Image,
-				)
+				output.Infof(localizer.Sprintf("Stopping %s", endpoint.ContainerDetails.Image))
```
shueybubbles, line 119: >of [](http://example.com/codeflow?start=14&length=2)

what happens if endpoint.ContainerDetails.Image contains a combination that also happens to be a Go format specifier? Won't that cause output.Infof to treat it as a format string?

```go
sub := "%d"
m := localizer.Sprintf("something %s", sub)
output.Infof(m)
```

shueybubbles: Go libraries typically have 2 versions of string functions - one with "f" and one without. Does `output` have `Info` and `Warn` functions ? If we aren't passing a format string we shouldn't be using and `f` function.

shueybubbles (changes requested):
:clock1:

cmd/modern/root/uninstall.go
apoorvdeshmukh: Yes. It could be an issue if sub contains format specifier. We have added these functions under output package. I'll add the non `f` variants `output.Info` and `output.Warn` and use them.

internal/output/output.go
```diff
@@ -223,3 +244,9 @@ func (o Output) printf(format string, a ...any) {
 	_, err := o.standardWriteCloser.Write([]byte(text))
 	o.errorCallback(err)
```
shueybubbles, line 245: printf should probably just call print(text)

shueybubbles:
> func (o Output) Errorf(format string, a ...any) {

can you update the linter to look for calls to Output functions that could lead to this bug in the future?
Today the linter doesn't block the build but at least someone running build\build locally could see the violation.

---
Refers to: internal/output/output.go:36 in 567c8e9. [](commit_id = 567c8e954c951be3625d6a325eb8a8f41dbd6928, deletion_comment = False)

cmd/modern/root/install/mssql-base.go
```diff
@@ -603,7 +603,7 @@ func (c *MssqlBase) downloadImage(
 	output *output.Output,
 	controller *container.Controller,
 ) {
-	output.Infof(localizer.Sprintf("Downloading %v", imageName))
+	output.Info(localizer.Sprintf("Downloading %v", imageName))
 	err := controller.EnsureImage(imageName)
 	if err != nil || c.unitTesting {
 		output.FatalfErrorWithHints(
```
shueybubbles, line 609: >.FatalfErrorWithHints [](http://example.com/codeflow?start=8&length=21)

this call looks problematic as well

shueybubbles:
> 		output.FatalfWithHints(

this call is fine because there are no user-provided arguments, but please check other callers of these Fatalf methods.

---
Refers to: cmd/modern/root/install/mssql-base.go:627 in 567c8e9. [](commit_id = 567c8e954c951be3625d6a325eb8a8f41dbd6928, deletion_comment = False)

internal/output/output.go
apoorvdeshmukh: Fixed

apoorvdeshmukh:
> > func (o Output) Errorf(format string, a ...any) {
> 
> can you update the linter to look for calls to Output functions that could lead to this bug in the future? Today the linter doesn't block the build but at least someone running build\build locally could see the violation.
> 
> Refers to: internal/output/output.go:36 in [567c8e9](https://github.com/microsoft/go-sqlcmd/commit/567c8e954c951be3625d6a325eb8a8f41dbd6928). [](commit_id = [567c8e9](https://github.com/microsoft/go-sqlcmd/commit/567c8e954c951be3625d6a325eb8a8f41dbd6928), deletion_comment = False)

I'll take this as a follow up item. Since we are using localization routines in all of these functions (except few places) , I have added the non f variants of the existing print functions and have replaced it everywhere.

cmd/modern/root/install/mssql-base.go
apoorvdeshmukh: Yes. Replaced the call with `FatalErrorWithHints`

apoorvdeshmukh:
> 

There were similar issues with other `f` functions too. Fixed them.

apoorvdeshmukh:
> > ```
> >   output.FatalfWithHints(
> > ```
> 
> this call is fine because there are no user-provided arguments, but please check other callers of these Fatalf methods.
> 
> Refers to: cmd/modern/root/install/mssql-base.go:627 in [567c8e9](https://github.com/microsoft/go-sqlcmd/commit/567c8e954c951be3625d6a325eb8a8f41dbd6928). [](commit_id = [567c8e9](https://github.com/microsoft/go-sqlcmd/commit/567c8e954c951be3625d6a325eb8a8f41dbd6928), deletion_comment = False)

There were similar issues with other `f` functions too. Fixed them.

shueybubbles (approved):
:shipit:

PR #418, opened 2023-07-08
Fix mac builds

apoorvdeshmukh:
Mac builds failed with following error
```
# github.com/microsoft/go-sqlcmd/cmd/modern/root/open
root/open/ads_darwin.go:31:14: undefined: localizer
##[error]The Go task failed with an error: Error: The process '/Users/runner/hostedtoolcache/go/1.18.0/x64/bin/go' failed with exit code 2
```
VS Code did not throw errors/warning as it warned that this file is within mdule "." which is not included in workspace.


shueybubbles (approved):
:shipit:

PR #440, opened 2023-07-31
Fix: space delimited args for -i and -v

### LINKED ISSUE #348: preserve compatibility with odbc sqlcmd for -v
specify multiple var=values values" is not right any more for -v.  
I think it should be "use the -v flag multiple times to pass in several values"

shueybubbles:
Fixes #348 
Short of completely rewriting the command line parser and ignoring Cobra, we are subject to a fair number of limitations for backward compatibility.

Of note:
- The `os.Args` array doesn't give any clue whether the argument was passed in quotes.
- Cobra doesn't parse the raw command line, it relies on the args array and its limitations.

This fix increases the number of command line variants that will be ODBC compatible but is far from exhaustive.
Users who don't use commas and dashes in their file names will be fine using space-delimited arguments for `-i`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,9 @@ The `sqlcmd` project aims to be a complete port of the original ODBC sqlcmd to t
 
 ### Changes in behavior from the ODBC based sqlcmd
 
+- `/` is not accepted as a flag specifier, only `-`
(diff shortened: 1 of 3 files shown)


apoorvdeshmukh (approved):
Is the limitation of os.Args not indicating the quotes, windows specific? 
The testcase with filename containing "-i" was quite insightful.
Thanks for updating the readme regarding comma limitations.

PR #441, opened 2023-07-31
Fix: Use proper reader for long file lines

### LINKED ISSUE #437: Can't handle long strings
What should I do with bufio.scanner: token too long? Can you use a unbuffered reader?

shueybubbles:
Fixes #437

### DIFF
--- a/pkg/sqlcmd/sqlcmd.go
+++ b/pkg/sqlcmd/sqlcmd.go
@@ -327,24 +327,26 @@ func (s *Sqlcmd) IncludeFile(path string, processAll bool) error {
 	b := s.batch.batchline
 	utf16bom := unicode.BOMOverride(unicode.UTF8.NewDecoder())
 	unicodeReader := transform.NewReader(f, utf16bom)
-	scanner := bufio.NewScanner(unicodeReader)
-	buf := make([]byte, maxLineBuffer)
-	scanner.Buffer(buf, maxLineBuffer)
+	scanner := bufio.NewReader(unicodeReader)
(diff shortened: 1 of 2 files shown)


pkg/sqlcmd/sqlcmd.go
```diff
@@ -327,24 +327,26 @@ func (s *Sqlcmd) IncludeFile(path string, processAll bool) error {
...
-		if !scanner.Scan() {
-			err := scanner.Err()
-			if err == nil {
-				return "", io.EOF
-			}
-			return "", err
+		var (
+			isPrefix bool  = true
+			err      error = nil
+			line, ln []byte
```
shueybubbles, line 337: > ln  [](http://example.com/codeflow?start=8&length=4)

i will experiment with pre-allocating ln as a property of s so we don't have to perform so many allocations.
There's no race condition so we should be able to reuse it.
The question becomes - is there a line length where we wouldn't want to keep the buffer allocated at that size? EG if we hit some line size like 1 GB, is it ok to keep a 1GB buffer around for the remainder of the process lifetime?

apoorvdeshmukh: You are freeing up the buffer for reuse anyway. I think the garbage collector would decide if it should be around for the process lifetime.

shueybubbles: setting the reference to `[:0]` doesn't free the buffer for GC, it just sets the length to 0 without reducing capacity.
The current code keeps the buffer around for the lifetime of the file read then lets it GC.

PR #447, opened 2023-08-08
Feat: Add -e and -k, finish -X, -r, -L

shueybubbles:
The flags with optional values require extra work to inject default values into the args array because Cobra doesn't handle flags with optional parameters.


apoorvdeshmukh (approved):
:shipit:

PR #452, opened 2023-08-16
Add product name and description to output

rwestMSFT:
Add product name to `--version` and `-?` output, to emulate the output of sqlcmd ODBC.

### DIFF
--- a/cmd/sqlcmd/sqlcmd.go
+++ b/cmd/sqlcmd/sqlcmd.go
@@ -224,6 +224,7 @@ func Execute(version string) {
 			setVars(vars, &args)
 
 			if args.Version {
+				fmt.Println(localizer.Sprintf("sqlcmd: Install/Create/Query SQL Server, Azure SQL, and Tools"))
 				fmt.Printf("%v\n", version)
 				fmt.Println(localizer.Sprintf("Legal docs and information: aka.ms/SqlcmdLegal"))
 				fmt.Println(localizer.Sprintf("Third party notices: aka.ms/SqlcmdNotices"))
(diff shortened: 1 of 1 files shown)


shueybubbles:
thx for opening a PR!

I don't think this particular output is relevant for `cmd/sqlcmd`, since legacy mode only supports queries. None of the flags -? prints are about install or create.
ODBC is printing the product name and we don't really have a good product name other than "sqlcmd" yet. @dlevy-msft it'd be good to talk to the marketing/branding folks to see what name we should use.

rwestMSFT:
@shueybubbles This is less about relevancy, and more about adding the name of the product when checking which version is installed. As discussed in other conversations, `sqlcmd "-?"` is a universal command that will run across OSes and versions of `sqlcmd`, and I think it should include a product name (whatever that is) before the version number, in this and `--version` info.

cmd/sqlcmd/sqlcmd.go
```diff
@@ -228,6 +228,7 @@ func Execute(version string) {
 			setVars(vars, &args)
 
 			if args.Version {
+				fmt.Println(localizer.Sprintf("sqlcmd: Install/Create/Query SQL Server, Azure SQL, and Tools"))
```
shueybubbles, line 231: >ocalizer.Sprintf("sqlcmd: Install/Create/Query SQL Server, Azure SQL, and Tools" [](http://example.com/codeflow?start=17&length=80)

2 things, for localizability
1 - declare the string in a higher scope so there's only 1 copy of it.
2 - run `build\build` on your Windows enlistment to generate the localization files to include in the PR

Without 1 I think the loc tools will have an error about a duplicate key.

rwestMSFT: @shueybubbles I'm not a Go developer by any stretch of the imagination, and this edit was done in-browser. I have no local environment to make this change.

stuartpa (approved):

cmd/sqlcmd/sqlcmd.go
stuartpa: Pushed above changes to branch

Alyst85:
✌️👍

PR #461, opened 2023-09-14
feat: Change Password

### LINKED ISSUE #458: Implement change password
- Add `-z` and -`Z` switches
- On connection failure check for error 18488. If the user didn't provide -P/-Z/-z, prompt for the new password

shueybubbles:
Fixes #458 
Like the old sqlcmd, `-Z` and `-z` are not mutually exclusive, but `-Z` trumps `-z`
If the user doesn't provide `-P`, `-z`, or '-Z' and the interactive connection fails with error 18488 we prompt the user for a new password.

### DIFF
--- a/cmd/sqlcmd/sqlcmd.go
+++ b/cmd/sqlcmd/sqlcmd.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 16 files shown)


apoorvdeshmukh (approved):
:shipit:

PR #495, opened 2024-01-02
Add initial telelmetry skeleton

apoorvdeshmukh:
- Telemetry and logging based on env variable

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/docker/go-connections v0.4.0
 	github.com/golang-sql/sqlexp v0.1.0
 	github.com/google/uuid v1.3.0
+	github.com/microsoft/ApplicationInsights-Go v0.4.4
 	github.com/microsoft/go-mssqldb v1.6.0
 	github.com/opencontainers/image-spec v1.0.2
 	github.com/peterh/liner v1.2.2
@@ -25,6 +26,7 @@ require (
 )
 
 require (
+	code.cloudfoundry.org/clock v0.0.0-20180518195852-02e53af36e6c // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.7.1 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.0 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.3.0 // indirect
@@ -40,6 +42,7 @@ require (
 	github.com/docker/go-units v0.5.0 // indirect
 	github.com/docker/libtrust v0.0.0-20160708172513-aabc10ec26b7 // indirect
 	github.com/fsnotify/fsnotify v1.6.0 // indirect
+	github.com/gofrs/uuid v3.3.0+incompatible // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-jwt/jwt/v5 v5.0.0 // indirect
 	github.com/golang-sql/civil v0.0.0-20220223132316-b832511892a9 // indirect
--- a/internal/telemetry/appinsights.go
+++ b/internal/telemetry/appinsights.go
@@ -1,0 +1,75 @@
+package telemetry
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/microsoft/ApplicationInsights-Go/appinsights"
+)
+
+var TelemetryClient appinsights.TelemetryClient
+
+func init() {
+	telemetryEnabled := strings.ToLower(os.Getenv("SQLCMD_TELEMETRY"))
+	if telemetryEnabled == "true" {
+		InitializeAppInsights()
+	}
+}
+
+func InitializeTelemetryLogging() {
+	telemetryEnabled := strings.ToLower(os.Getenv("SQLCMD_TELEMETRY"))
+	loggingEnabled := strings.ToLower(os.Getenv("SQLCMD_TELEMETRY_LOGGING"))
+	if telemetryEnabled == "true" && loggingEnabled == "true" {
+		// Add a diagnostics listener for printing telemetry messages
+		appinsights.NewDiagnosticsMessageListener(func(msg string) error {
+			fmt.Printf("[%s] %s\n", time.Now().Format(time.UnixDate), msg)
+			return nil
+		})
+	}
+}
+
+func InitializeAppInsights() {
+	instrumentationKey := "f305b208-557d-4fba-bf06-25345c4dfdbc"
+	config := appinsights.NewTelemetryConfiguration(instrumentationKey)
+	TelemetryClient = appinsights.NewTelemetryClientFromConfig(config)
+	InitializeTelemetryLogging()
+}
+
+func TrackEvent(eventName string, properties map[string]string) {
+	event := appinsights.NewEventTelemetry(eventName)
+	for key, value := range properties {
+		event.Properties[key] = value
+	}
+	TelemetryClient.Track(event)
+}
+
+func CloseTelemetry() {
+	select {
+	case <-TelemetryClient.Channel().Close(10 * time.Second):
+		// Ten second timeout for retries.
+
+		// If we got here, then all telemetry was submitted
+		// successfully, and we can proceed to exiting.
+	case <-time.After(30 * time.Second):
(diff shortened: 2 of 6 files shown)


AxzelBC (commented):
It is feasible to use Viper for environment variables. But, it looks good. 👍

go.mod
```diff
@@ -5,11 +5,13 @@ go 1.21
 require (
 	github.com/alecthomas/chroma/v2 v2.5.0
 	github.com/billgraziano/dpapi v0.4.0
+	github.com/denisbrodbeck/machineid v1.0.1
```
shueybubbles, line 8: >machineid v1.0.1 [](http://example.com/codeflow?start=26&length=16)

should have a NOTICE update too

cmd/modern/root/config/add-context.go
```diff
@@ -93,4 +94,23 @@ func (c *AddContext) run() {
 		{localizer.Sprintf("To start interactive query session"), "sqlcmd query"},
 		{localizer.Sprintf("To run a query"), "sqlcmd query \"SELECT @@version\""},
 	}, localizer.Sprintf("Current Context '%v'", context.Name))
+	c.LogTelemtry()
+}
+
+func (c *AddContext) LogTelemtry() {
```
shueybubbles, line 100: >emtry( [](http://example.com/codeflow?start=27&length=6)

Telemetry

cmd/modern/root/config/add-context.go
```diff
@@ -93,4 +94,23 @@ func (c *AddContext) run() {
 		{localizer.Sprintf("To start interactive query session"), "sqlcmd query"},
 		{localizer.Sprintf("To run a query"), "sqlcmd query \"SELECT @@version\""},
 	}, localizer.Sprintf("Current Context '%v'", context.Name))
+	c.LogTelemtry()
```
shueybubbles, line 97: >c.LogTelemtry() [](http://example.com/codeflow?start=1&length=15)

instead of requiring `run` implementation to call this directly, is it possible to define an interface with `LogTelemetry` as a member and have whatever calls `run` also call it if the object implements it?

cmd/modern/root/config/add-context.go
```diff
@@ -93,4 +94,23 @@ func (c *AddContext) run() {
...
+}
+
+func (c *AddContext) LogTelemtry() {
+	eventName := "config-add-context"
+	properties := map[string]string{}
+	properties["Command"] = "config"
+	properties["SubCommand"] = "add-context"
+	if c.name != "" {
+		properties["name"] = "set"
+	}
```
shueybubbles, line 107: This function looks mostly like boilerplate that could be implemented in some common place and just have the command implementation pass the values not the property names for Command and SubCommand.

It's probably also useful to define a new type like

```Go
type CommandEvent map[string]string

func (e *CommandEvent) NewCommandEvent(command string, subCommand string) {
// Set Command and SubCommand properties on e
}

func (e *CommandEvent) SetProperty(p string, v string) *CommandEvent {
// Add the given property to the map and return e
}
```

cmd/modern/root/config/add-user.go
```diff
@@ -170,3 +171,23 @@ func (c *AddUser) run() {
 	uniqueUserName := config.AddUser(user)
 	output.Info(localizer.Sprintf("User '%v' added", uniqueUserName))
 }
+
+func (c *AddUser) LogTelemtry() {
+	eventName := "config-add-user"
+	properties := map[string]string{}
+	if c.username != "" {
+		properties["username"] = "set"
```
shueybubbles, line 179: is it easier to build telemetry reports if a property always has a value instead of omitting it when not used? What do the report queries look like?

internal/cmdparser/interface.go
```diff
@@ -42,4 +42,8 @@ type Command interface {
 	// the Command object to have access to the dependencies it needs, without
 	// having to manage them directly.
 	SetCrossCuttingConcerns(dependency.Options)
+
+	// LogTelemtry is used to track the useful telemetry for the command
+	// To be enforced after the review
+	// LogTelemtry()
```
shueybubbles, line 48: >// LogTelemtry() [](http://example.com/codeflow?start=1&length=16)

if we switch to a new method, it should probably pass the map as a parameter and not require the command to invoke telemetry methods directly.

internal/telemetry/appinsights.go
```diff
@@ -0,0 +1,302 @@
...
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/microsoft/ApplicationInsights-Go/appinsights"
+	"github.com/microsoft/go-sqlcmd/internal/localizer"
+)
+
+var telemetryClient appinsights.TelemetryClient
+var isTelemetryEnabled string = "true"
```
shueybubbles, line 16: >d string = "true" [](http://example.com/codeflow?start=21&length=17)

it's odd to use a string as a bool

cmd/modern/root/config/add-context.go
```diff
@@ -93,4 +94,23 @@ func (c *AddContext) run() {
 		{localizer.Sprintf("To start interactive query session"), "sqlcmd query"},
 		{localizer.Sprintf("To run a query"), "sqlcmd query \"SELECT @@version\""},
 	}, localizer.Sprintf("Current Context '%v'", context.Name))
+	c.LogTelemtry()
+}
+
+func (c *AddContext) LogTelemtry() {
+	eventName := "config-add-context"
```
shueybubbles, line 101: >eventName := "config-add-context" [](http://example.com/codeflow?start=1&length=33)

How does this `eventName` differ from the event name constants defined in internal/telemetry/appinsights.go?

internal/telemetry/appinsights.go
```diff
@@ -0,0 +1,302 @@
...
+
+const (
+
+	//Common Properties
+	UserId    = "Sqlcmd.userId"
+	SessionId = "Sqlcmd.sessionId"
+	LocLang   = "Sqlcmd.locLang"
+	UserOs    = "Sqlcmd.userOs"
+
+	// Legacy Flags
```
shueybubbles, line 68: >// Legacy Flags [](http://example.com/codeflow?start=1&length=15)

putting these as constants implies maintenance overhead every time we add a new flag.

Is there any reason not to build these identifiers dynamically based on the cobra data? In which scenario will an engineer have to type these constant symbols into a code editor?

internal/telemetry/appinsights.go
```diff
@@ -0,0 +1,302 @@
...
+		// Add a diagnostics listener for printing telemetry messages
+		appinsights.NewDiagnosticsMessageListener(func(msg string) error {
+			fmt.Printf("[%s] %s\n", time.Now().Format(time.UnixDate), msg)
+			return nil
+		})
+	}
+}
+
+func InitializeAppInsights() {
+	instrumentationKey := "f305b208-557d-4fba-bf06-25345c4dfdbc"
```
shueybubbles, line 260: >"f305b208-557d-4fba-bf06-25345c4dfdbc" [](http://example.com/codeflow?start=23&length=38)

can we parameterize this key at build time? 
It may be desirable to have one key for local builds and a separate production key for our signed builds.

internal/telemetry/userid_windows.go
```diff
@@ -0,0 +1,29 @@
...
+func getOrCreateUserId() string {
+
+	k, err := registry.OpenKey(registry.CURRENT_USER, `Software\Microsoft\SQMClient`, registry.ALL_ACCESS)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer k.Close()
+
+	s, _, err := k.GetStringValue("UserId")
+	if err != nil && strings.Contains(err.Error(), "The system cannot find the file specified") {
```
shueybubbles, line 20: > "The system cannot find the file specified") { [](http://example.com/codeflow?start=47&length=47)

this won't work on non-English Windows

internal/telemetry/userid_windows.go
```diff
@@ -0,0 +1,29 @@
...
+		id := uuid.New()
+		k.SetStringValue("UserId", "{"+strings.ToUpper(id.String()+"}"))
+		s, _, _ = k.GetStringValue("UserId")
+	}
+	if strings.Contains(s, "{") {
+		return s[1 : len(s)-1]
```
shueybubbles, line 26: >return [](http://example.com/codeflow?start=2&length=6)

s =

PR #498, opened 2024-01-04
Backward compat support for -N

### LINKED ISSUE #497: sqlcmd -N should work the same as -N true
For backward compat we can treat `-N` by itself as `true`

shueybubbles:
treat `-N` by itself as `-N true`
Fixes #497


apoorvdeshmukh (approved):
:shipit:

PR #543, opened 2024-07-23
fix: Parse list flags correctly

shueybubbles:
Handle multiple `-i` and `-v` entries correctly when they are followed by a flag-without-space like `-isomefile1 somefile2 -V10`


dlevy-msft (approved):

apoorvdeshmukh (approved):
:shipit:

PR #559, opened 2024-12-12
Fix: Make colon optional for reset

### LINKED ISSUE #552: 'reset' command not supported
The ODBC version of `sqlcmd `supports the `reset `command, which discards the batch (as opposed to `go`, which sends the batch to the server).
The Go version of `sqlcmd `does not seem to support reset, but my scripts use it. Please add support for `reset`?

shueybubbles:
Fixes #552


apoorvdeshmukh (approved):
:shipit:

PR #589, opened 2025-06-24
Feat: Add HostNameInCertificate for strict encryption

### LINKED ISSUE #504: Need a way to pass the HostNameInCertificate parameter for strict encryption connections
For legacy mode we have a few options:
1. Add a new flag for HNIC. 
2. Make HNIC an optional parameter to `-C`, such that `-C X` means "trust the cert only if its subject is X"
3. Expand the optional arguments to `-N` so that each encryption type can have parameters of its own.
Something like this:
`-N strict:HNIC=myhost.com`

Option 1 is a pain because we are running out of flags. Option 2 solves for this specific problem.

Option 3 is intriguing because it opens the door for supporting more encryption parameters in the future without needing to add new flags. It also simplifies detection of invalid flag combinations.

@dlevy-msft 

@stuartpa how does the modern mode allow the user to set encryption values for connections to non-container instances?

shueybubbles:
Fixes #504 
Since ODBC sqlcmd uses `-F` for the host name, we are going to follow suit and switch to `--vertical` as the parameter to turn on vertical output formatting.

`sqlcmd -S someserver -Nstrict -F *.mydomain.com`


shueybubbles:
i will generate new localization files after merging the driver update change when it goes to main, just to avoid dealing with merge conflicts.

PR #708, opened 2026-02-09
fix: invalid handling of empty values in sqlcmd

debarshibasak:
The built sqlcmd is throwing panic errors, for the empty values. For eg.
```
SQLCMDPASSWORD=<somepassword> ./sqlcmd -S <ip> -U 'sa' -d '' -Q "SELECT name FROM sys.databases WHERE name NOT IN ('tempdb','sys','msdb', 'model')" -b
```

Throws:

```
runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/microsoft/go-sqlcmd/cmd/sqlcmd.checkDefaultValue({0xc000034150, 0x9, 0xefcf38?}, 0x4)
	/__w/1/s/go-sqlcmd/cmd/sqlcmd/sqlcmd.go:320 +0x2da
github.com/microsoft/go-sqlcmd/cmd/sqlcmd.convertOsArgs({0xc000034150, 0x9, 0x9})
	/__w/1/s/go-sqlcmd/cmd/sqlcmd/sqlcmd.go:302 +0x22a
github.com/microsoft/go-sqlcmd/cmd/sqlcmd.Execute({0x122eaac, 0x6})
	/__w/1/s/go-sqlcmd/cmd/sqlcmd/sqlcmd.go:274 +0x150
main.main()
	/__w/1/s/go-sqlcmd/cmd/modern/main.go:52 +0x209
```

After the fix:

name                                                                                                                            
--------------------------------------------------------------------------------------------------------------------------------
pubs

I have validated for rest of the flags too.


debarshibasak:
> ```
> @microsoft-github-policy-service agree
> ```

@microsoft-github-policy-service agree

dlevy-msft-sql:
Fixes #750

dlevy-msft-sql:
Thanks for the fix @debarshibasak and apologies for the long delay getting CI unblocked on this one! The root-cause analysis is spot on, and routing both call sites through `isFlag` instead of repeating `args[i+1][0] == '-'` is a nice cleanup. Tightening `assert.ElementsMatch` → `assert.Equal` in the test is also a real win, since arg order is the whole contract of `convertOsArgs`.

Before we merge, could you make a couple of small tweaks?

### 1. Rebase onto current `main`

The branch is a few months behind and `main` has since moved to Go 1.25.10 plus a few CI changes. A rebase + force-push will also re-trigger the workflows cleanly.

### 2. Add a test for the `__complete ""` panic

Issue #750 reports the same crash via a different path: `sqlcmd __complete ''` (empty arg that isn't preceded by a single-char flag). Your fix already covers it, but it would be great to lock that in with a regression test so it can't sneak back. Something like:

```go
{
    "empty arg standalone (issue #750)",
    []string{"__complete", ""},
    []string{"__complete", ""},
},
{
    "minimal -d empty (issue #708)",
    []string{"-d", ""},
    []string{"-d", ""},
},
```

The minimal `-d ""` case is also worth having alongside the longer one. It's easier to read as "this is the exact panic from the issue."

### 3. Pin the behavior for `-N ""` and friends

Before this change, `-N ""` panicked. After it, `isFlag("")` returns false, so `checkDefaultValue` no longer injects `"true"` for `-N ""` (and similarly no default for `-r ""`, `-k ""`, `-L ""`, `-X ""`). That's a reasonable choice, empty string isn't really "missing value", but it's a behavior change worth a test so a future refactor doesn't flip it accidentally:

```go
{
    "-N with empty next arg",
    []string{"-N", ""},
    []string{"-N", ""}, // adjust if you'd prefer "true" to be appended
},
```

Whichever behavior you pick is fine with us; we just want it nailed down in the table.

Once those are in and CI is green, this is good to merge. Thanks again for tracking this down and for your patience on the review!

PR #731, opened 2026-04-10
ci: add dependabot grouping and conventional commit instructions

dlevy-msft-sql:
- Group all gomod updates into a single PR per week
- Group all github-actions updates into a single PR per week
- Add conventional commit format guide to copilot-instructions.md

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -227,6 +227,100 @@ go generate ./...
 4. Add tests for new functionality
 5. Keep commits focused and well-described
 
+## Commit Message Format
+
+This project uses [Conventional Commits](https://www.conventionalcommits.org/) for automated version management and changelog generation via [Release Please](https://github.com/googleapis/release-please).
+
+### Required Format
+
+**ALWAYS** use conventional commit format for PR titles and commit messages:
+
+```
+<type>: <description>
+
+[optional body]
+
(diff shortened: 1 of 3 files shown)


.github/copilot-instructions.md
```diff
@@ -227,6 +227,100 @@ go generate ./...
 4. Add tests for new functionality
 5. Keep commits focused and well-described
 
+## Commit Message Format
+
+This project uses [Conventional Commits](https://www.conventionalcommits.org/) for automated version management and changelog generation via [Release Please](https://github.com/googleapis/release-please).
```
shueybubbles, line 232: is there any go workflow that would auto-reject a PR that doesn't use this approach?

shueybubbles (approved):

.github/copilot-instructions.md
dlevy-msft-sql: No, it just wouldn't show up in the list. It would still merge and ride along with that release. 

Should we add something to prevent those stealth releases?

dlevy-msft-sql: @shueybubbles I just pushed a commit to enforce the naming convention. Stealth releases just don't feel right. The rule checks that the proper prefix is used and that it is in lower case.

shueybubbles (approved):

PR #747, opened 2026-04-30
fix: reject non-YAML extensions for --sqlconfig flag

### LINKED ISSUE #597: Add documentation/help about supported config types
Hi, sorry for my poor English.

I'm a newer to this sqlcmd utility, version 1.8.2, trying to use configuration file to manage connection contexts.

Seems that sqlcmd _sometimes_ use the configuration file's extension to determine which "config type" to use.
After some rough tests, seems recognized config types are:
- .yaml
- .yml
- .toml
- .json
- .ini
- (maybe others?)

Using other file extensions, like .sqlconfig, running command like
```sh
sqlcmd config --sqlconfig ./myproject.sqlconfig add-endpoint --name ep1434 --address localhost --port 1434
```
sqlcmd will complain:
> Error: Unsupported Config Type "sqlconfig"

where _sqlconfig_ seems to be the file extension.

However, commands that read (rather than write) configuration file seem to treat the configuration file as YAML-formatted,
e.g.
```sh
sqlcmd config --sqlconfig ./myproject.sqlconfig view
```

I'm not sure if I'm missing something, but so far I cannot find somewhere documented this behavior.
Or maybe there's a plan adding a list of supported config type / a guide of naming a configuration file?

### LINKED ISSUE #690: Add validation/warning for unsupported config file extensions
**Description:**
Currently, when users specify a config file with an unsupported extension (e.g., `--sqlconfig myfile.txt`), they receive a confusing Viper error: "Unsupported Config Type 'txt'".

**Proposed Enhancement:**
Add validation in `internal/config/viper.go` to check the file extension before passing to Viper and provide a clearer error message such as:
```
Error: Configuration files must use YAML format with .yaml or .yml extension.
The file 'myfile.txt' has an unsupported extension '.txt'.
```

**Related:**
- Issue #597 - Original documentation request
- PR #684 - Documents YAML-only requirement

**Implementation Notes:**
- Check file extension in `configureViper()` or `SetFileName()` function
- Allow `.yaml`, `.yml`, and no extension (for default `~/.sqlcmd/sqlconfig`)
- Provide helpful error message pointing users to documentation

dlevy-msft-sql:
## Problem

`sqlcmd --sqlconfig config.json` silently creates a JSON-named file but writes YAML content, leading to confusing behavior when users reference it later.

Fixes #690, Fixes #597

## Root Cause

No extension validation in `config.SetFileName()`.

## Solution

- Validate file extension in `SetFileName()` before proceeding
- Reject extensions other than `.yaml` or `.yml` (extensionless files like the default `~/.sqlcmd/sqlconfig` are still allowed)
- Report via `checkErr` (consistent with existing error patterns in this package)

## Changes

| File | Change |
|------|--------|
| `internal/config/config.go` | Extension validation in `SetFileName` |
| `internal/config/config_test.go` | Tests for rejected and accepted extensions |
| `README.md` | Document custom config file extension requirement |

## Testing

- `go build ./cmd/modern` passes
- `go test ./internal/config/...` passes
- New `TestSetFileName_RejectsNonYAMLExtension` verifies `.json`/`.toml` rejected and `.yaml`/`.yml` accepted

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -4,13 +4,16 @@
 package config
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
 	. "github.com/microsoft/go-sqlcmd/cmd/modern/sqlconfig"
 	"github.com/microsoft/go-sqlcmd/internal/io/file"
 	"github.com/microsoft/go-sqlcmd/internal/io/folder"
 	"github.com/microsoft/go-sqlcmd/internal/pal"
-	"os"
-	"path/filepath"
-	"testing"
 )
 
 var config Sqlconfig
@@ -22,6 +25,11 @@ var filename string
 func SetFileName(name string) {
 	if name == "" {
 		panic("name is empty")
+	}
+
+	if ext := strings.ToLower(filepath.Ext(name)); ext != "" && ext != ".yaml" && ext != ".yml" {
+		checkErr(fmt.Errorf("configuration file %q has unsupported extension %q (must be .yaml or .yml)", name, ext))
+		return
 	}
 
 	filename = name
--- a/README.md
+++ b/README.md
@@ -84,6 +84,15 @@ sqlcmd config connection-strings
 sqlcmd config view
 ```
 
+#### Custom configuration files
+
+The `--sqlconfig` flag specifies a custom configuration file. The file must be YAML. Extensionless filenames are allowed; if the file has an extension, it must be `.yaml` or `.yml`:
+
+```
+sqlcmd config --sqlconfig ./myproject.yaml add-endpoint --name ep1434 --address localhost --port 1434
+sqlcmd config --sqlconfig ./myproject.yaml view
+```
+
 ### Versions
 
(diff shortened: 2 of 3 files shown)


internal/config/config.go
```diff
@@ -4,13 +4,16 @@
 package config
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
```
shueybubbles, line 11: why do non-test files import `testing` ?

shueybubbles (approved):

internal/config/config.go
dlevy-msft-sql: Good catch, but this isn't introduced by this PR — `SetFileNameForTest(t *testing.T)` (and the `testing` import) was added in commit `10ee238` ("Modern CLI - Part 2"); it only appears in this diff because gofmt regrouped the import block. Same pattern exists in `internal/cmdparser/test.go`.

Filed #780 to track the cleanup separately, since the fix touches files outside the scope of this PR.

PR #811, opened 2026-09-14
fix: keep the Msg/Level/State header on errors sent to stderr with -r

### LINKED ISSUE #794: -r discards the "Msg N, Level N, State N" error header — errors on stderr lose message number and line info
### Description

When `-r` (errors to stderr) is specified, server error messages are written to stderr
as **bare message text only** — the `Msg %d, Level %d, State %d, Server %s, Line %d`
header line is silently dropped. Without `-r`, the header prints correctly (to stdout).

ODBC sqlcmd with `-r` prints the *same* formatted output (header + message), just
redirected to stderr. So scripts/CI pipelines migrating from ODBC sqlcmd lose the error
number and line information exactly in the configuration meant for error capture.

### Repro

```sql
-- repro.sql
SELECT id, badColumn FROM sys.objects WHERE object_id = 1;
GO
```

```
> sqlcmd -S myserver -E -b -i repro.sql          # no -r
Msg 207, Level 16, State 1, Server MYSERVER, Line 1
Invalid column name 'id'.

> sqlcmd -S myserver -E -r -b -i repro.sql       # with -r
Invalid column name 'id.'        <-- stderr: message text only, no Msg/Level/State/Line
```

ODBC sqlcmd (16.0.1000.6) with the identical `-r -b` flags writes to stderr:

```
Msg 207, Level 16, State 1, Server MYSERVER, Line 1
Invalid column name 'id'.
Msg 207, Level 16, State 1, Server MYSERVER, Line 1
Invalid column name 'badColumn'.
```

### Root cause (from source)

`Formatter.AddError` in `pkg/sqlcmd/format.go` formats the header correctly for
`mssql.Error`. But when `-r` is passed, `cmd/sqlcmd/sqlcmd.go` installs a
`s.PrintError` hook that receives only `e.Message` and writes it directly to
`os.Stderr`, returning `true` — which **bypasses `Format.AddError` entirely**:

```go
s.PrintError = func(msg string, severity uint8) bool {
    if severity >= stderrSeverity {
        s.WriteError(os.Stderr, errors.New(msg+sqlcmd.SqlcmdEol))
        return true
    }
    return false
}
```

Suggested fix: format the full header inside the hook (or route through the formatter
with a stderr destination) so `-r` only changes the destination, not the format —
matching ODBC sqlcmd behavior.

### Additional observation

Combined with `-b`, only the **first** error of a batch is reported (the message loop
exits on the first qualifying error), while ODBC sqlcmd prints all errors of the
response before exiting. Together with the missing header this significantly degrades
error diagnostics for deployment scripts.

### Environment

- sqlcmd v1.10.0 (sqlcmd-windows-amd64.zip from Releases)
- Windows 11, SQL Server 2022 (16.0.1190.2)
- Compared against ODBC sqlcmd 16.0.1000.6 with identical flags
- (Server messages in our environment are localized/Korean; behavior is language-independent)

Raymondriter:
Fixes #794

## Problem

With `-r`, `runQuery` passed only `e.Message` to the `PrintError` hook. Because the hook returns `true`, `Formatter.AddError` was skipped, and that's where the `Msg N, Level N, State N, Server S, Line N` header is built. Errors on stderr lost the error number and line. ODBC sqlcmd prints the same text with `-r` and only changes the destination.

## Fix

- Move the header formatting from `AddError` into `errorHeader(e mssql.Error)`. The localized format strings are unchanged, so existing translations still apply.
- Pass `errorHeader(e) + e.Message` to `PrintError`, so a host that redirects errors gets the text the default output would show.
- The `PrintError` signature is unchanged. I documented on the field that `msg` includes the header for server errors.

## Tests

- New `TestPrintErrorIncludesErrorHeader` runs `RAISERROR (N'Testing!', 11, 1)` with a redirecting hook. It asserts the hook receives `Msg 50000, Level 11, State 1, Server ..., Line 1` followed by the message, and that nothing reaches the default output. On `main` the hook receives only `Testing!`.
- Full suite against a local SQL Server 2022 container: the same results as `main`. The only failures are the environment-dependent install/ADS/editor tests.
- End to end with the repro from the issue (`sqlcmd -r -b -i repro.sql`), stderr now reads:

  ```
  Msg 207, Level 16, State 1, Server <host>, Line 1
  Invalid column name 'id'.
  ```

  This matches the output without `-r`.

Not addressed here: the issue's additional observation that `-b` stops after the first error of a batch. It's a separate behaviour change, and I can follow up in its own PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Raymondriter:
Withdrawing this for now. The root cause and fix are described above in case they're useful to anyone picking up #794.

